PR #4, opened 2026-03-29
Fix MemoryPromptSectionBuilder contract for OpenClaw 2026.3.28

fuller-stack-dev:
## Summary
- Fixes `TypeError: memorySection is not iterable` crash on OpenClaw 2026.3.28 caused by breaking change in the `MemoryPromptSectionBuilder` contract
- The builder now returns `string[]` (prompt lines) and accepts `{ availableTools, citationsMode }` instead of the previous `{ id, content }` return / `{ userId, messages }` args
- Per-query vector recall continues to work via the context engine's `assemble()` hook — the prompt section is now a lightweight static header

## Context
OpenClaw 2026.3.28 changed how `buildMemorySection` consumes the registered prompt builder. The gateway spreads the return value (`...memorySection`) into the system prompt array, so a non-iterable object crashes the agent before it can reply.

Two errors in the gateway log:
1. `TypeError: memorySection is not iterable` — spread of `{ id, content }` object
2. `TypeError: Cannot read properties of undefined (reading 'at')` — `args.messages` is now undefined since the new params are `{ availableTools, citationsMode }`

## Test plan
- [x] Unit tests updated and passing (`test/unit/memory-provider.test.ts`)
- [ ] Integration test with OpenClaw 2026.3.28 gateway — verify agent boots without error and memory section appears in system prompt

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -21,7 +21,7 @@ export default definePluginEntry({
     api.registerContextEngine("libravdb-memory", () =>
       buildContextEngineFactory(runtime.getRpc, cfg, recallCache),
     );
-    api.registerMemoryPromptSection(buildMemoryPromptSection(runtime.getRpc, cfg, recallCache));
+    api.registerMemoryPromptSection(buildMemoryPromptSection());
     api.on("gateway_stop", () => runtime.shutdown());
   },
 });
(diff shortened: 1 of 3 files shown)


xDarkicex (approved):
Safe to merge. The PR fixes a real OpenClaw 2026.3.28 contract break in registerMemoryPromptSection. It does trade away the old recall-cache prewarm, so there may be a small retrieval-latency regression, but I don’t see a correctness or safety blocker.

PR #18, opened 2026-04-08
fix: add missing context engine .info property and auto-discover daemon

fuller-stack-dev:
Three bugs that interact to crash the OpenClaw gateway when the libravdb-memory context engine is active:

1. buildContextEngineFactory() returns an object without an .info property. The gateway accesses params.contextEngine.info.id to check for the legacy engine, causing a TypeError on every agent run. Fix: add a full info: { id, name, ownsCompaction } object to the returned engine.

2. ownsCompaction was set as a top-level property on the engine object, but the gateway reads it exclusively from contextEngine.info.ownsCompaction. This meant compaction ownership was silently ignored. Fix: move ownsCompaction: true into the info object (kept top-level for backward compat with any internal callers).

3. defaultEndpoint() hardcodes ~/.clawdb/run/libravdb.sock, but the Homebrew formula LaunchAgent places the socket at /opt/homebrew/var/clawdb/run/. Users hit ENOENT on first connect and silently fall into degraded mode. Fix: check LIBRAVDB_RPC_ENDPOINT env var first, then probe well-known paths before falling back to the original user-local path.

## Summary by CodeRabbit

* **Chores**
  * Updated internal metadata for the LibraVDB Memory context engine to include additional identification and capability information.

### DIFF
--- a/src/context-engine.ts
+++ b/src/context-engine.ts
@@ -59,7 +59,7 @@ export function buildContextEngineFactory(
   }
 
   return {
-    info: { id: "libravdb-memory" },
+    info: { id: "libravdb-memory", name: "LibraVDB Memory", ownsCompaction: true },
     ownsCompaction: true,
     async bootstrap({ sessionId, userId }: ContextBootstrapArgs) {
       const rpc = await getRpc();


xDarkicex (approved):
looks great, thanks fuller-stack_dev

PR #28, opened 2026-04-21
fix: keep libravdb sidecar transport binary

fuller-stack-dev:
## Summary
- stop defaulting the LibraVDB sidecar supervisor socket to UTF-8 so protobuf RPC traffic stays binary
- add regression coverage in the sidecar lifecycle test harness by simulating Node socket string mode when `setEncoding()` is enabled
- fix the install-time failure mode where `libravdb-memory` could load but then break session turns before compaction because RPC frames arrived as strings

## Testing
- `pnpm run build`
- `pnpm exec tsc -p tsconfig.tests.json && mkdir -p .ts-build/src/generated && cp -rf src/generated/. .ts-build/src/generated/ && node --test .ts-build/test/unit/rpc.test.js .ts-build/test/integration/sidecar-lifecycle.test.js`

## Notes
- `pnpm run check` is currently failing on `origin/main` because `scripts/setup.ts` imports `./child-env.js` without declarations; this PR does not change that existing issue.

## Summary by CodeRabbit

* **New Features**
  * Plugin now supports multiple kind categories for enhanced classification.

* **Improvements**
  * Startup health check timeout now dynamically calculated based on RPC configuration for better reliability.
  * Refined socket encoding behavior for more flexible data handling.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -14,7 +14,7 @@ export default definePluginEntry({
   id: "libravdb-memory",
   name: "LibraVDB Memory",
   description: "Persistent vector memory with three-tier hybrid scoring",
-  kind: "context-engine",
+  kind: ["memory", "context-engine"],
 
   register(api: OpenClawPluginApi) {
     const cfg = api.pluginConfig as PluginConfig;
--- a/openclaw.plugin.json
+++ b/openclaw.plugin.json
@@ -3,7 +3,7 @@
   "name": "LibraVDB Memory",
   "description": "Persistent vector memory with three-tier hybrid scoring",
   "version": "1.4.9",
-  "kind": "context-engine",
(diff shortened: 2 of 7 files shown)


src/index.ts
```diff
@@ -14,7 +14,7 @@ export default definePluginEntry({
   id: "libravdb-memory",
   name: "LibraVDB Memory",
   description: "Persistent vector memory with three-tier hybrid scoring",
-  kind: "context-engine",
+  kind: ["memory", "context-engine"],
```
xDarkicex, line 17: Good

xDarkicex:
The actual issue is that setEncoding('utf8') switches sockets to string mode, corrupting binary protobuf frames, not that 'utf8 causing tool call crashes'. Consider updating the PR description.

PR #29, opened 2026-04-21
fix: map budget compaction requests to legacy sidecar fields

fuller-stack-dev:
## Summary
- map OpenClaw's newer budget-style `contextEngine.compact()` calls onto LibraVDB's older `compact_session` wire contract by deriving `targetSize` from `tokenBudget`
- forward continuity compaction settings that the sidecar already understands during explicit compaction requests
- add focused host-flow coverage for the compatibility mapping so overflow/timeout retry requests keep working against the current sidecar API

## Testing
- `./node_modules/.bin/tsc -p tsconfig.tests.json`
- `./node_modules/.bin/tsc -p tsconfig.build.json --pretty false`
- `node --test .ts-build/test/integration/host-flow.test.js`
- live runtime smoke test after swapping the built `dist/context-engine.js` into the installed extension and restarting the gateway: a patched throwaway session stayed healthy through oversized follow-up probes and continued normal tool/assistant cycles instead of immediately reproducing the old `400 Your input exceeds the context window of this model` loop

## Notes
- `pnpm check` is currently blocked by an unrelated existing upstream type issue in `scripts/setup.ts` (`child-env.js` declaration missing), so I used the narrower compile/test path above.

## Summary by CodeRabbit

* **New Features**
  - Extended session compaction API to accept optional token budget parameter.
  - Extended session state update API to support optional token budget and runtime context parameters.

### DIFF
--- a/src/context-engine.ts
+++ b/src/context-engine.ts
@@ -20,6 +20,33 @@ export function buildContextEngineFactory(
   recallCache: RecallCache<SearchResult>,
   logger: LoggerLike = console,
 ) {
+  function buildCompactSessionRequest(args: {
+    sessionId: string;
+    force?: boolean;
+    targetSize?: number;
+    tokenBudget?: number;
+  }): Partial<CompactSessionRequest> {
+    // OpenClaw core now requests budget-style compaction using tokenBudget,
+    // but the current LibraVDB compact_session wire contract still expects
+    // targetSize. Use tokenBudget as the compatibility target so overflow and
+    // timeout retries still compact toward the host's requested prompt budget.
+    const targetSize = args.targetSize ?? args.tokenBudget;
+    return {
+      sessionId: args.sessionId,
+      force: args.force,
+      ...(typeof targetSize === "number" ? { targetSize } : {}),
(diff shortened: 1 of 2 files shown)


xDarkicex:
"The mapping const targetSize = args.targetSize ?? args.tokenBudget; is a bit aggressive because it treats a token budget as a turn count.

Since libravdbd uses targetSize to determine cluster density, passing a large number like 2048 (tokens) will cause the daemon to collapse the entire session into a single giant cluster.

Recommendation: Consider either:

Dividing the budget by a constant (e.g., Math.floor(args.tokenBudget / 400)) to get a realistic turn count.
Updating the compact_session RPC in the daemon to accept a native tokenBudget field so the daemon can use its internal token-counting logic."

xDarkicex (commented):
Safe to merge: this is a compatibility shim that maps host compaction budget requests onto the legacy daemon targetSize field, so the daemon contract stays unchanged while OpenClaw’s newer budget-based flow works correctly.

xDarkicex (approved):
Changes match spec

PR #32, opened 2026-04-22
fix: wait for sidecar reconnect before rpc send

fuller-stack-dev:
## Summary
- wait for the sidecar supervisor socket to reconnect when a transient `Sidecar socket unavailable` gap happens between long-running turns
- retry the pending RPC send once after reconnect instead of surfacing the temporary socket gap immediately
- add a focused RPC unit test that simulates the reconnect window and verifies the original call succeeds after reconnect

## Testing
- `pnpm build`
- `pnpm exec tsc -p tsconfig.tests.json && mkdir -p .ts-build/src/generated && cp -rf src/generated/. .ts-build/src/generated/ && node --test .ts-build/test/unit/rpc.test.js`

## Context
This PR is intentionally separate from the already-merged adapter normalization work in #30. It addresses the next remaining plugin-side issue observed in long monitored compaction runs: transient sidecar reconnect gaps could still abort a turn with `Sidecar socket unavailable` even when the daemon came back quickly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved RPC client resilience with automatic reconnection retry handling for temporary socket unavailability issues.

### DIFF
--- a/src/rpc.ts
+++ b/src/rpc.ts
@@ -37,36 +37,66 @@ export class RpcClient {
     return await new Promise<T>((resolve, reject) => {
       const id = ++this.seq;
       const timeoutMs = callOptions.timeoutMs ?? this.options.timeoutMs;
+      const deadline = Date.now() + timeoutMs;
       const timer = setTimeout(() => {
         this.pending.delete(id);
         reject(new Error(`RPC timeout: ${method} (${timeoutMs}ms)`));
       }, timeoutMs);
 
       this.pending.set(id, { resolve, reject, timer, decodeResult: codec.decodeResult });
-      const envelope = new (RpcRequest as any)({
-        id,
-        method,
-        params: codec.encodeParams(params),
-      });
-      const payload = Buffer.from(envelope.toBinary());
-      const header = Buffer.alloc(4);
-      header.writeUInt32BE(payload.byteLength, 0);
 
-      const chunks: Buffer[] = [];
-      if (!this.sentMagic) {
-        chunks.push(Buffer.from([0x02]));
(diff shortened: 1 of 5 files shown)


xDarkicex:
P1
Reconnect retry can send a stale RPC after the caller has already timed out
Dismiss
When socket.write(frame) throws, waitForReconnect(timeoutMs).then(() => send(false)) is detached from the original call timer. If the sidecar reconnects after the RPC has already timed out, the promise is rejected but the queued send(false) still fires later, which can deliver a stale ingest or after_turn request after the caller has already seen a timeout. That retry needs to be canceled or gated on the pending call still being active.

fuller-stack-dev:
Addressed the reconnect review feedback in fe02fea.

Changes made:
- rebuild the RPC frame on each send attempt so magic-byte state is recomputed after reconnect
- only flip sentMagic after a successful write
- gate reconnect retries on pending.has(id) so calls removed by timeout or close are never resent later
- use a deadline-based remaining timeout for reconnect waits
- add off() to the sidecar socket interface and clean up complementary reconnect listeners on settle
- strengthen the reconnect unit coverage to include prior traffic, post-close reconnect, and timeout-no-resend behavior

Validation rerun:
- pnpm build
- pnpm exec tsc -p tsconfig.tests.json && mkdir -p .ts-build/src/generated && cp -rf src/generated/. .ts-build/src/generated/ && node --test .ts-build/test/unit/rpc.test.js

xDarkicex:
Congratulation’s approved.

xDarkicex (approved):
Looks good my friend

PR #38, opened 2026-04-27
fix: enable LibraVDB CLI in discovery mode

fuller-stack-dev:
## Summary
- create the LibraVDB plugin runtime in `discovery` mode so lazy-loaded `openclaw memory` subcommands register real handlers on newer OpenClaw builds
- keep memory and context-engine activation gated to `full` mode only, so discovery stays non-activating outside CLI command registration
- add a regression test covering discovery-mode CLI registration

## Testing
- pnpm run test:ts
- npm run build
- openclaw memory status
- openclaw memory journal --limit 1
- openclaw memory search --query foo

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -19,11 +19,14 @@ export function register(api: OpenClawPluginApi) {
     return;
   }
 
-  const isFullMode = (api.registrationMode as string) === "full";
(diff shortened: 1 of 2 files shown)


fuller-stack-dev:
Superseded by #39, which targets main.

PR #41, opened 2026-04-27
fix: truncate mtimeMs to integer before BigInt conversion

compoodment:
## Problem

Node.js `fs.stat().mtimeMs` returns a float (e.g. `1777104808165.547`). The daemon RPC protobuf layer converts `sourceMtimeMs` to `BigInt`, which rejects non-integer values:

```
The number 1777104808165.547 cannot be converted to a BigInt because it is not an integer
```

This breaks **all** markdown ingestion on Linux/WSL2 where filesystem mtimes commonly have sub-millisecond precision. Every workspace `.md` file fails to ingest at startup.

## Fix

`Math.trunc()` both the call site (`stat.mtimeMs`) and the RPC params field (`sourceMtimeMs`) to produce integer milliseconds before BigInt conversion.

## Testing

- TypeScript compiles clean (`tsc --noEmit` passes)
- Patched plugin running on WSL2 with daemon v1.4.25 — all 7 workspace files ingested successfully, database grew from 12KB to 119KB
- No unit test changes needed (no behavioral change beyond the type coercion)

## Reproduction

On any Linux filesystem with sub-millisecond mtime precision, start the gateway with `markdownIngestionRoots` configured and watch the daemon logs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed timestamp precision handling in the markdown ingestion pipeline to ensure consistent integer-based formatting of file modification times.

### DIFF
--- a/src/markdown-ingest.ts
+++ b/src/markdown-ingest.ts
@@ -434,7 +434,7 @@ class DirectoryMarkdownSourceAdapter implements MarkdownSourceAdapter {
       this.fileStates.delete(sourceDoc);
       return;
     }
-    await this.ingestMarkdownDocument(sourceDoc, text, rootState.root, relativePath, fileHash, stat.size, Math.trunc(stat.mtimeMs));
+    await this.ingestMarkdownDocument(sourceDoc, text, rootState.root, relativePath, fileHash, stat.size, stat.mtimeMs);
     this.fileStates.set(sourceDoc, {
       root: rootState.root,
       sourceDoc,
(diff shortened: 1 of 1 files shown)


xDarkicex (approved):
Nice catch — thanks for tracking this down. The float-to-BigInt crash on Linux/WSL2 is exactly the kind of cross-platform edge case that's easy to miss. `Math.trunc` at the RPC boundary is the right fix.

Also pushed a minor follow-up: dropped the redundant `Math.trunc` at the call site since `ingestMarkdownDocument` already guards the boundary. Approved and good to merge.
Thank you @compoodment

PR #43, opened 2026-04-27
feat: plugin-side identity resolution for cross-session CLI memory re…

xDarkicex:
…call

OpenClaw's ContextEngine interface (v2026.4.11+) does not pass userId to plugin hooks — the concept was removed during the plugin system revamp. Without userId, the namespace resolver fell back to ephemeral sessionKey, making every CLI invocation a separate memory namespace.

This adds plugin-side identity resolution:
- identity.ts: resolves stable userId from plugin config → identity JSON file → OS auto-derive (username@hostname#homeHash) → sessionKey fallback
- memory-scopes.ts: replaces durable-namespace.ts with scope-based retrieval routing (session/user/global) and crossSessionRecall gating
- context-engine.ts: resolves identity once, caches it, passes derived userId to all daemon RPC calls (bootstrap, ingest, assemble, afterTurn)
- memory-runtime.ts: identity resolution wired into the memory search path
- types.ts: new config fields userId, identityPath, crossSessionRecall

Identity file defaults to $OPENCLAW_STATE_DIR/libravdb-identity.json (falls back to ~/.openclaw/libravdb-identity.json).

## Summary by CodeRabbit

* **New Features**
  * Configurable user identity management with stable cross-session identities and new settings: userId, identityPath, crossSessionRecall
  * Memory retrieval scopes expanded to session/user/global with durable user namespaces

* **Improvements**
  * Structured logging for lifecycle entrypoints and safer dispatch with failure logging
  * Registration flow now better distinguishes lightweight modes and logs details

* **Tests**
  * New integration/unit tests for cross-session durable memory; memory-related assertions now validate collection shapes rather than exact strings


fuller-stack-dev:
Runtime-path suggestion after local validation:

I think this PR should explicitly account for every OpenClaw path that can create an agent turn, not just the direct CLI path. In local testing, `openclaw agent --session-id ...` did not increase LibraVDB durable counts or make a planted marker searchable, while a Gateway `sessions.patch` + `sessions.send` run did ingest the marker and a separate session later recalled it. That means the real contract to validate is: every supported entrypoint must reach the same context-engine lifecycle (`bootstrap`/`assemble`/`afterTurn` or ingest fallback) with a stable `sessionId`, `sessionKey`, and durable `userId`.

Suggested acceptance matrix:

1. Gateway / Control UI path: create or patch a session, send through `sessions.send`, assert `afterTurn` runs and the marker is visible through `openclaw memory search` under the durable user namespace.
2. Channel path: simulate a channel-backed inbound message, e.g. Discord/webchat-style `deliveryContext`, assert the same stable identity is used and the turn is durably searchable.
3. Direct CLI path: either make `openclaw agent --session-id` use the same embedded-runner/context-engine lifecycle, or document it as intentionally outside durable-memory coverage and avoid using it as a memory validation path.
4. Cron/background/session-target path: verify scheduled or system-originated turns still provide the same `sessionKey`/identity metadata and do not fall back to session-local-only memory.
5. Subagent/spawned session path: verify parent/child sessions preserve durable user scope while still isolating session-local recall.

Implementation suggestion:

- Add a small shared runtime conformance test helper that instruments a fake context engine and records calls to `bootstrap`, `assemble`, `afterTurn`, `ingestBatch`, and `ingest`.
- Drive each entrypoint above through its normal public surface and assert the final lifecycle call includes non-empty `sessionId`, expected `sessionKey`, and the same resolved durable user identity.
- Add one live-ish LibraVDB smoke script/test that plants a unique marker through the Gateway/channel-like path, waits until `openclaw memory search --agent <agent> --query <marker>` returns it, then asks a separate session to recall it. The check should read the persisted session transcript too, because `sessions.send --expect-final` can return before the final assistant text is visible in immediate CLI output.
- Add explicit negative diagnostics for skipped ingestion: when a turn completes but is not ingested, log whether it was missing lifecycle invocation, missing session identity, gated out by `ingestionGateThreshold`, or delayed by async indexing.

Why this matters:

The local result showed durable memory can work on the Gateway path, but ingestion was path-sensitive and somewhat timing-sensitive. This PR is about plugin-side identity resolution, so the safest bar is not just "search works once"; it is "all normal OpenClaw turn producers converge on the same durable identity and lifecycle hooks."

PR #45, opened 2026-04-29
fix(cli): make memory index a compatibility no-op

fuller-stack-dev:
## Summary
- make `openclaw memory index` a true compatibility no-op for LibraVDB instead of spinning up the runtime and doing redundant status RPCs
- update the CLI help and docs to clarify that LibraVDB keeps index state inside the sidecar
- add a unit test that locks in the no-runtime-start behavior for the index command

## Why
- operators and retry loops still invoke `openclaw memory index --force` for generic OpenClaw compatibility, but LibraVDB does not have a host-side reindex pass to run
- the previous implementation still started the plugin runtime and hit the sidecar just to print a delegated message, which adds avoidable churn and log noise under repeated retries

## Testing
- `pnpm run test:ts`
- `pnpm run build`

## Summary by CodeRabbit

* **Documentation**
  * Updated documentation to clarify that `openclaw memory index --force` is retained for backward compatibility but no longer performs a reindex operation.

* **Refactor**
  * The memory index command now executes as a compatibility no-op, with index state managed internally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,10 @@ openclaw memory journal --limit 50
 openclaw memory dream-promote --user-id <userId> --dream-file /path/to/DREAMS.md
 ```
 
+`openclaw memory index --force` is kept as a compatibility command. LibraVDB
+maintains its own index state inside the sidecar, so the command exits without
+starting an explicit reindex pass.
+
 Use [Install](./docs/install.md) for daemon lifecycle commands and
 [Uninstall](./docs/uninstall.md) for safe shutdown and removal.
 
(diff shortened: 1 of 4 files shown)


compoodment (approved):
Reviewed the diff and tests. This looks correct: memory index is now a real compatibility no-op, no sidecar/runtime startup happens, docs match the behavior, and the new test covers the key invariant. Approving.

xDarkicex:
Appreciate the PR, but no-opping a broken integration point isn't the direction we're taking. When the plugin surface doesn't map to a real daemon capability, the fix is to add the missing capability — which is now in progress. Closing in favor of wiring runIndex to the new rebuild_index daemon RPC.

PR #46, opened 2026-04-29
fix: stop leaking prePromptMessageCount to daemon in afterTurn

xDarkicex:
## Summary
- `afterTurn` was spreading `...args` into the RPC call, leaking `prePromptMessageCount`, `tokenBudget`, and `runtimeContext` to the daemon
- The daemon uses `prePromptMessageCount` to skip messages by array position — if the framework passes a stale value, messages are silently dropped before persistence
- The daemon already has content-hash dedup (`afterTurnIngestedKeys`), so the positional skip is redundant for correctness

## Fix
- Kernel path: removed `prePromptMessageCount` forwarding
- RPC path: replaced `...args` spread with explicit `{ sessionId, sessionKey, userId, messages, isHeartbeat }`
- Both paths now let the daemon default `prePromptMessageCount` to 0, processing all messages; content-hash dedup catches any already-ingested

## Test plan
- [x] `npm run test:integration` — all 30 pass
- [x] `npm run test:ts` — 72 pass, 10 pre-existing failures (unrelated, present on clean main)

## Summary by CodeRabbit

* **Refactoring**
  * Optimized internal messaging to reduce unnecessary data transmission in RPC communications.

### DIFF
--- a/src/context-engine.ts
+++ b/src/context-engine.ts
@@ -978,7 +978,6 @@ export function buildContextEngineFactory(
             sessionKey: args.sessionKey,
             userId,
             messages,
-            prePromptMessageCount: args.prePromptMessageCount,
             isHeartbeat: args.isHeartbeat,
           });
           await performAfterTurnPredictiveCompaction({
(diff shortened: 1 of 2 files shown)


compoodment (approved):
Reviewed the kernel and RPC paths. This looks correct: the RPC call now forwards only daemon-relevant fields, stripping prePromptMessageCount/tokenBudget/runtimeContext, and the kernel path no longer sends prePromptMessageCount. The updated test covers the key invariant. Minor future cleanup: mark/remove the proto field when compatibility allows. Approving.

xDarkicex:
@compoodment Thank you for checking, merging now.

PR #50, opened 2026-04-29
fix: remove unused markdown collection option

compoodment:
## Summary

Removes the documented/schema `markdownIngestionCollection` option because it is not actually wired to the daemon markdown ingestion RPC.

## Why

The option appears in:

- `openclaw.plugin.json`
- `PluginConfig`
- `docs/features.md`

…but no runtime code reads it. More importantly, `IngestMarkdownDocumentRequest` has no `collection` field, and `MarkdownSourceMeta` has no target collection field either. So users can set `markdownIngestionCollection`, but it cannot affect where markdown is stored.

Leaving it documented is misleading. Until the daemon/proto supports target collections for authored markdown, the honest fix is to remove the dead config surface.

## Evidence

```text
grep -R "markdownIngestionCollection" src test docs openclaw.plugin.json
```

Only type/schema/docs references exist; no ingest path uses it.

Generated IPC shape:

- `IngestMarkdownDocumentRequest`: `sourceDoc`, `text`, `tokenizerId`, `coreDoc`, `sourceMeta`
- `MarkdownSourceMeta`: source/provenance/hash fields only

## Tests

```text
npx tsc --noEmit
corepack pnpm run test:ts -- --test-name-pattern="markdown ingestion|manifest"
```

Unit suite passes. Full integration on main is currently blocked by the brittle checklist assertion fixed separately in #48.

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -51,7 +51,6 @@ Relevant config fields:
 | `markdownIngestionRoots` | Generic markdown roots to watch. |
 | `markdownIngestionInclude` | Optional include globs for generic roots. |
 | `markdownIngestionExclude` | Optional exclude globs for generic roots. |
-| `markdownIngestionCollection` | Target collection for generic markdown, default `global`. |
 | `markdownIngestionDebounceMs` | Watch debounce window, default `150`. |
 | `markdownIngestionObsidianEnabled` | Enables Obsidian ingestion when vault roots exist. |
 | `markdownIngestionObsidianRoots` | Obsidian vault roots to watch. |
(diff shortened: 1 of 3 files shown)


xDarkicex:
Confirmed safe to merge. The removed `markdownIngestionCollection` config field had zero runtime references — checked the proto
  (`IngestMarkdownDocumentRequest` has no `collection` field), the RPC call site in `src/markdown-ingest.ts`, and the full repo grep. Build passes
   cleanly.

  Also confirmed with the daemon side: the `authored:*` collections are tied to the assembly engine's deontic routing (hard/soft/variant tiers),
  so a user-facing collection override on ingest would bypass that system. This was dead config, not a missing feature.
  
  Thank you @compoodment

PR #51, opened 2026-04-29
feat: wire index command to daemon rebuild_index RPC

xDarkicex:
## Summary
- Replaces the stubbed `runIndex` with a real call to the daemon's `rebuild_index` RPC
- `--force` is required — refuses to run without it
- Namespace resolved from `--user-id`, `--session-key`, or `--agent`
- Optional `--collections` flag limits scope to specific collections
- Prints results: collections processed, records reindexed, collections recreated

## Test plan
- [x] `npm run test:ts` — 84 pass, 0 fail
- [x] Type check clean (`tsc --noEmit`)

## Summary by CodeRabbit

* **New Features**
  * Added scoping parameters (--user-id, --session-key) and optional --collections to the memory index command.
  * Status checks can now trigger the explicit rebuild path for the memory index.

* **Changes**
  * The memory index workflow is now an explicit rebuild flow and requires --force.
  * Rebuild runs report processing/reindex/recreate counts, warn on errors, and set non-zero exit codes when failures occur.


compoodment:
Two concrete blockers in this PR:

**1. Missing protobuf codec for `rebuild_index`**

`rpc.call("rebuild_index", ...)` will fail before reaching the daemon. The plugin's codec registry (`src/rpc-protobuf-codecs.ts`) has no `rebuild_index` entry, and the generated protobuf bindings (`src/generated/libravdb/ipc/v1/rpc_pb.ts`) contain no `RebuildIndexRequest`/`RebuildIndexResponse` message types. The proto schema (`api/proto/intelligence_kernel/v1/kernel.proto`) also has no `rebuild_index` RPC.

The call will hit `getRpcMethodCodec` → `undefined` → encode failure with: `Unsupported LibraVDB RPC method for protobuf transport: rebuild_index`. Even if the daemon supports `rebuild_index` internally, the plugin can't serialize/deserialize it without the proto types and codec.

Fix path: add proto message definitions, regenerate bindings, add the codec entry, add a test.

**2. `status --index` path is now broken**

The old `runIndex` did a lightweight sync and didn't require `--force`. The new `runIndex` requires `--force`, but `status` calls `runIndex(runtime, cfg, { ...opts, verbose: false }, logger, { quiet: true })` without `--force`. So `openclaw memory status --index` will always produce the rebuild-required error instead of refreshing status.

Fix path: give `status --index` its own lightweight sync path (like the old bridge-based one), remove `--index` from `status` since it no longer does what it advertised, or add `--force` to `status` and update the wording.

**Also worth noting:** one-shot CLI commands currently start markdown ingestion and then shut the runtime down while scans are still in flight (#53, #54). Since this PR changes the CLI `index` path, it may be worth handling shutdown cleanly in the same pass or avoiding long-lived markdown ingestion in CLI mode.

src/cli.ts
```diff
@@ -292,41 +296,53 @@ async function runStatus(
...
-    await manager.sync?.({
-      reason: "cli",
-      force: Boolean(opts?.force),
+    const rpc = await runtime.getRpc();
+    const result = await rpc.call<{
+      collectionsProcessed: number;
+      recordsReindexed: number;
+      collectionsRecreated: number;
+      errors: string[];
+    }>("rebuild_index", {
```
fuller-stack-dev, line 318: `RpcClient.call()` rejects methods that are not present in `rpcProtobufCodecs` before sending anything to the daemon. This PR adds a `rebuild_index` call, but no codec/generated request-response mapping is added for that method, so `memory index --force` will fail with `Unsupported LibraVDB RPC method for protobuf transport: rebuild_index`. Please add the protobuf codec/types for this RPC or call an existing supported method.

PR #54, opened 2026-04-30
fix: stop markdown ingestion before CLI runtime shutdown

### LINKED ISSUE #53: CLI memory commands race markdown ingestion during shutdown
## Summary

Running LibraVDB CLI commands such as `openclaw memory status`, `index`, or `search` can print markdown ingestion failures during CLI shutdown even when the daemon is healthy and the command itself succeeds.

This looks like the CLI invocation starts the full plugin runtime, including background markdown ingestion, then `runCliCommand()` shuts the runtime down while the startup scan is still in progress.

## Repro environment

- OpenClaw: 2026.4.23
- `openclaw-memory-libravdb`: 1.4.22
- `libravdbd`: 1.4.26
- WSL2/Linux x86_64
- Libra isolated profile using `plugins.slots.memory = "libravdb-memory"`
- External daemon active on `unix:/home/computment/.clawdb/run/libravdb.sock`
- `LIBRAVDB_ONNX_DEVICE=cpu`

## Repro

```bash
OPENCLAW_STATE_DIR=$HOME/.openclaw-Libra \
OPENCLAW_CONFIG_PATH=$HOME/.openclaw-Libra/openclaw.json \
OPENCLAW_GATEWAY_PORT=19001 \
openclaw memory status
```

## Observed

The command succeeds and reports healthy status:

```text
Sidecar           running
Turns stored      1
Memories stored   1
Embedding profile all-minilm-l6-v2
Message           ok
```

But then emits shutdown/race errors:

```text
[plugins] [markdown-ingest] sync failed for .../workspace-Libra/AGENTS.md: Socket closed
[plugins] [markdown-ingest] sync failed for .../workspace-Libra/HEARTBEAT.md: LibraVDB plugin runtime has been shut down
[plugins] [markdown-ingest] sync failed for .../workspace-Libra/IDENTITY.md: LibraVDB plugin runtime has been shut down
...
```

Same pattern reproduced with:

```bash
openclaw memory index --verbose
openclaw memory search --query 'memory plugin stress tester' --json
```

## Expected

CLI commands should not start long-lived markdown ingestion, or they should wait/cancel it cleanly before shutdown. Successful CLI status/search/index commands should not emit scary ingestion failure logs.

## Notes

Persistent gateway startup in the isolated test profile does not show this error once the daemon is running. Direct daemon checks pass:

- `status` -> `ok`
- `ingest_markdown_document` -> `ok`
- `after_turn_kernel` -> `ok`
- `assemble_context_internal` recalls both authored docs and session turns

So this appears isolated to one-shot CLI plugin runtime lifecycle, not daemon health or embedding failure.

compoodment:
## Summary

Fixes #53 by making plugin-owned background tasks participate in runtime shutdown.

The CLI commands call `runtime.shutdown()` after the command action. Before this change, markdown ingestion and dream promotion were only stopped from the `gateway_stop` hook, so one-shot CLI invocations could start markdown ingestion and then close the runtime/socket while the startup scan was still in flight. That produced misleading shutdown errors such as `Socket closed` and `LibraVDB plugin runtime has been shut down` even when the daemon was healthy and the CLI command succeeded.

## Changes

- Add `PluginRuntime.onShutdown()` for plugin-owned background cleanup.
- Register markdown ingestion and dream promotion cleanup with the runtime shutdown path.
- Keep `gateway_stop` cleanup routed through `runtime.shutdown()` so gateway and CLI share one shutdown path.
- Make markdown ingestion `stop()` wait for active scans before resolving, canceling timers/watchers first so no new scans are scheduled.
- Suppress per-file sync warnings during intentional markdown ingestion shutdown.
- Add coverage proving markdown ingestion stop waits for an in-flight startup scan.

## Tests

```text
PATH=/tmp/corepack-bin:$PATH pnpm run check
npm run test:integration
```

Results:

```text
pnpm run check: 82/82 unit tests passed
npm run test:integration: 31/31 integration tests passed
```

## Release Notes

* **New Features**
  * Added shutdown hook mechanism for coordinated component lifecycle management.

* **Bug Fixes**
  * Improved shutdown process to reliably wait for in-flight markdown scans to complete before stopping, preventing interrupted operations.
  * Enhanced shutdown coordination to prevent new work from being scheduled during the shutdown phase.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -95,8 +95,10 @@ export function register(api: OpenClawPluginApi) {
   const dreamPromotion = createDreamPromotionHandle(cfg, runtime.getRpc, api.logger ?? console);
 
   runtime.onShutdown(async () => {
-    await dreamPromotion.stop();
     await markdownIngestion.stop();
+  });
+  runtime.onShutdown(async () => {
+    await dreamPromotion.stop();
   });
 
   void markdownIngestion.start().catch((error) => {
--- a/src/index.ts
+++ b/src/index.ts
@@ -94,6 +94,11 @@ export function register(api: OpenClawPluginApi) {
   const markdownIngestion = createMarkdownIngestionHandle(cfg, runtime.getRpc, api.logger ?? console);
   const dreamPromotion = createDreamPromotionHandle(cfg, runtime.getRpc, api.logger ?? console);
 
+  runtime.onShutdown(async () => {
+    await dreamPromotion.stop();
+    await markdownIngestion.stop();
+  });
+
   void markdownIngestion.start().catch((error) => {
     api.logger?.warn?.(`LibraVDB markdown ingestion failed to start: ${error instanceof Error ? error.message : String(error)}`);
   });
@@ -104,8 +109,6 @@ export function register(api: OpenClawPluginApi) {
   api.on("before_reset", createBeforeResetHook(runtime, api.logger ?? console));
   api.on("session_end", createSessionEndHook(runtime, api.logger ?? console));
   api.on("gateway_stop", async () => {
-    await dreamPromotion.stop();
-    await markdownIngestion.stop();
     await runtime.shutdown();
   });
 }
--- a/src/markdown-ingest.ts
+++ b/src/markdown-ingest.ts
@@ -193,9 +193,11 @@ class DirectoryMarkdownSourceAdapter implements MarkdownSourceAdapter {
   private readonly logger: LoggerLike;
   private readonly states = new Map<string, RootState>();
   private readonly fileStates = new Map<string, FileState>();
+  private readonly activeScans = new Set<Promise<void>>();
   private readonly tokenizerId: string;
   private readonly coreDoc: boolean;
   private started = false;
+  private stopping = false;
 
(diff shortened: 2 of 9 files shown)


compoodment:
Reviewing the two CodeRabbit flags:

**1. Bundled shutdown tasks** — already addressed. The PR registers \`dreamPromotion.stop()\` and \`markdownIngestion.stop()\` as separate \`onShutdown\` callbacks (lines 97-98 and 100-101), so one rejection cannot prevent the other from running. Looks like the bot flagged an older diff state.

**2. \`stopping\` flag race** — false positive. The claim is that a late callback could enqueue work after \`stop()\` resolves because \`start()\` resets \`stopping = false\` before setting \`started = true\`. But \`scheduleRootScan()\` and \`scanRoot()\` both guard on \`!this.started || this.stopping\`, and \`scheduleRootScan\` returns early when \`!this.started\`. After \`stop()\` completes, \`started\` is \`false\` — so no timer callback or watcher event can enqueue new work regardless of the \`stopping\` flag state. The only path that sets \`stopping = false\` is \`start()\`, which then immediately sets \`started = true\` — and scheduling should resume after restart. No race window exists.

xDarkicex:
Thanks @compoodment — clean fix, good integration test proving the race is closed.

PR #56, opened 2026-04-30
fix: start runtime before selecting gRPC kernel

compoodment:
## Summary
- make `PluginRuntime.getKernel()` startup-backed so gRPC kernel transport is reachable on cold context-engine calls
- await kernel selection in context-engine lifecycle paths before falling back to the RPC sidecar transport
- remove the stale `RecallCache` plumbing that is no longer used by the synchronous memory prompt section or assembler

## Why
The context engine checked `runtime.getKernel()` synchronously before anything had started the plugin runtime. On a cold runtime this always returned `null`, so the first `bootstrap`, `assemble`, `ingest`, `compact`, or `afterTurn` call fell through to the RPC sidecar path even when `grpcEndpoint` was configured.

The recall cache was a leftover from the old prompt-section retrieval path. It was constructed and threaded through APIs, but never read or written by runtime code.

## Testing
- [x] `git diff --check`
- [x] `npm run test:ts` — 84/84 pass

## Summary by CodeRabbit

* **Refactor**
  * Removed the in-memory recall cache and its public types.
  * Kernel access is now asynchronous and failsafe, with graceful fallback when unavailable.
  * Simplified builder interfaces by removing the recall-cache parameter.

* **Tests**
  * Updated and expanded tests to exercise async kernel retrieval, fallback routing, and lifecycle flows.


xDarkicex:
Root cause is clear: getKernel() checked a started flag synchronously that was never set before the first caller, so the gRPC kernel path was
  permanently dead on cold runs. The fix awaits ensureStarted() inside getKernel(), which is the same async guard already used by getRpc() —
  consistent and correct.

  getKernelOrNull() properly guards all five lifecycle paths against rejection, and the dead RecallCache removal is clean with no dangling
  references.
  
  Thanks @compoodment

PR #57, opened 2026-04-30
fix: honor disabled cross-session recall in memory runtime

compoodment:
## Summary
- make the memory runtime search bridge respect `crossSessionRecall: false`
- limit searches to the active session collection when a session id is available
- return no runtime-search results instead of querying durable user/global collections when no session id is available
- add unit coverage for both paths

## Why
`crossSessionRecall` was already documented as disabling durable user recall and exact recall already honors it, but the memory runtime bridge ignored it. Calls through the memory search manager could still query `user:*` and `global` collections even when config requested session-only recall.

## Testing
- [x] `git diff --check`
- [x] `npm run test:ts` — 84/84 pass

### DIFF
--- a/src/memory-runtime.ts
+++ b/src/memory-runtime.ts
@@ -183,6 +183,9 @@ async function searchResolvedCollections(
   k: number,
 ): Promise<{ results: SearchResult[] }> {
   const collections = resolveSearchCollections(cfg, userId, sessionId);
+  if (collections.length === 0) {
+    return { results: [] };
+  }
   return collections.length === 1
     ? await rpc.call<{ results: SearchResult[] }>("search_text", {
         collection: collections[0],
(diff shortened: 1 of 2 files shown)


src/memory-runtime.ts
```diff
@@ -198,21 +201,27 @@ async function searchResolvedCollections(
 }
 
 function resolveSearchCollections(cfg: PluginConfig, userId: string, sessionId?: string): string[] {
+  if (cfg.crossSessionRecall === false) {
```
fuller-stack-dev, line 204: When `crossSessionRecall` is false, this only gates the non-dream path. A query that matches `detectDreamQuerySignal()` still takes the branch above and searches `dream:${userId}` even without a `sessionId`, which keeps durable cross-session recall enabled for dream queries. Please route/gate the dream path the same way or add a test documenting that dream recall is intentionally exempt.

compoodment:
Thanks @fuller-stack-dev for catching this. I've opened #61 to address the dream query bypass — when `crossSessionRecall === false`, dream queries now fall through to the session-scoped path instead of hitting `dream:${userId}`. Includes tests for both enabled and disabled cases.

xDarkicex:
honor disabled cross-session recall in memory runtime

  - make the memory runtime search bridge respect crossSessionRecall: false
  - limit searches to the active session collection when a session id is available
  - return no runtime-search results instead of querying durable user/global collections when no session id is available
  - add unit coverage for both paths

  Why:
  crossSessionRecall was already documented as disabling durable user recall and exact
  recall already honors it, but the memory runtime bridge ignored it. Calls through the
  memory search manager could still query user:* and global collections even when config
  requested session-only recall.

  Testing:
  - npm run test:ts — 84/84 pass
  - npm run test:integration — 30/30 pass

PR #58, opened 2026-04-30
fix: allow memory flush by session key

compoodment:
## Summary
- stop registering `memory flush --user-id` as a Commander required option
- keep runtime validation requiring either `--user-id` or `--session-key`
- add CLI registration coverage so `--session-key` remains usable while `dream-promote` keeps its truly required options

## Why
The docs advertise both:

```bash
openclaw memory flush --user-id <userId>
openclaw memory flush --session-key <sessionKey>
```

But when Commander supports `requiredOption`, the CLI registered `--user-id` as required. That means `memory flush --session-key <sessionKey>` would be rejected by the CLI parser before `runFlush()` could apply its intended “user id or session key” validation.

## Testing
- [x] `git diff --check`
- [x] `npm run test:ts` — 82/82 pass

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -147,11 +147,7 @@ export function registerMemoryCli(
 
       const flush = ensureCommand(root, "flush")
         .description("Wipe a durable memory namespace after confirmation");
-      if (flush.requiredOption) {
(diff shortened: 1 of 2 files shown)


xDarkicex:
Lekker bezig, @compoodment. Staat er strak in.

PR #59, opened 2026-04-30
fix: validate configured sidecar endpoints

compoodment:
## Summary
- tighten `sidecarPath` endpoint validation for configured daemon endpoints
- reject empty Unix socket endpoints and malformed TCP endpoints before socket creation
- add unit coverage for invalid endpoint shapes

## Why
`resolveConfiguredEndpoint()` only checked whether `sidecarPath` started with `unix:` or `tcp:`. Values like `unix:`, `tcp:`, `tcp:127.0.0.1`, `tcp:127.0.0.1:notaport`, or out-of-range TCP ports passed validation and failed later with lower-level socket/runtime errors.

This keeps the existing user-facing config error at the boundary where the invalid `sidecarPath` is parsed.

## Testing
- [x] `git diff --check`
- [x] `npm run test:ts` — 83/83 pass

### DIFF
--- a/src/sidecar.ts
+++ b/src/sidecar.ts
@@ -585,7 +585,21 @@ function describeEndpoint(endpoint: string): string {
 }
 
 function isConfiguredEndpoint(value?: string): boolean {
-  return value?.startsWith("tcp:") === true || value?.startsWith("unix:") === true;
+  if (!value) return false;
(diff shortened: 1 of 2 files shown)


xDarkicex:
Goeie aanvulling, @compoodment. Is gemerged.

PR #61, opened 2026-05-01
fix: gate dream query path behind crossSessionRecall config

compoodment:
Addresses the gap identified by @fuller-stack-dev on PR #57.

## Problem

When `crossSessionRecall === false`, the dream query branch in `search()` still routes to `dream:${userId}`, which is inherently cross-session (no session scoping). This bypasses the config gate that PR #57 added for the normal search path.

## Fix

- Changed the dream query condition from `dreamQuery.active` to `dreamQuery.active && cfg.crossSessionRecall !== false`. When cross-session recall is disabled, dream queries fall through to `searchResolvedCollections`, which already handles the `crossSessionRecall === false` case correctly.

- Tests cover enabled, disabled, and no-session-id cases for dream queries under crossSessionRecall.

All 10 tests pass.

## Notes

This targets `main` since cross-repo PRs cannot target a non-default branch. If #57 merges first, this can be rebased. The dream gate change is a single line and applies cleanly on top of #57.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed search and recall routing to properly respect the cross-session recall configuration setting.
  * Improved session-scoped search functionality: when cross-session recall is disabled, search is now correctly limited to the current session.
  * Enhanced safe handling of empty collection results.

* **Tests**
  * Extended test coverage for conditional search routing based on configuration settings.

### DIFF
--- a/src/memory-runtime.ts
+++ b/src/memory-runtime.ts
@@ -183,6 +183,9 @@ async function searchResolvedCollections(
   k: number,
 ): Promise<{ results: SearchResult[] }> {
   const collections = resolveSearchCollections(cfg, userId, sessionId);
+  if (collections.length === 0) {
+    return { results: [] };
+  }
   return collections.length === 1
     ? await rpc.call<{ results: SearchResult[] }>("search_text", {
         collection: collections[0],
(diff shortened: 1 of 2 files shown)


xDarkicex:
Safe. One line, correct fallthrough, well-tested, clean merge.

PR #72, opened 2026-05-03
fix: defer background services to gateway_start hook

xDarkicex:
## Summary
- Moves markdown ingestion and dream promotion `start()` calls from fire-and-forget during `register()` into a `gateway_start` hook
- Ensures background services start after the gateway is fully initialized, matching the documented plugin lifecycle
- The existing `gateway_stop` hook already handles graceful shutdown of both services

## Test plan
- [x] `gateway_stop` hook still calls `runtime.shutdown()` which drains `onShutdown` tasks (markdown/dream stop)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -119,11 +119,13 @@ export function register(api: OpenClawPluginApi) {
     await dreamPromotion.stop();
   });
 
-  void markdownIngestion.start().catch((error) => {
-    api.logger?.warn?.(`LibraVDB markdown ingestion failed to start: ${error instanceof Error ? error.message : String(error)}`);
-  });
-  void dreamPromotion.start().catch((error) => {
-    api.logger?.warn?.(`LibraVDB dream promotion failed to start: ${error instanceof Error ? error.message : String(error)}`);
+  api.on("gateway_start", async () => {
+    await markdownIngestion.start().catch((error) => {
+      api.logger?.warn?.(`LibraVDB markdown ingestion failed to start: ${error instanceof Error ? error.message : String(error)}`);
+    });
+    await dreamPromotion.start().catch((error) => {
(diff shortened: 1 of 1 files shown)


compoodment:
Heads up: this, #73, and #75 all modify adjacent regions of `src/index.ts`. If this merges after #73 or #75 (or vice versa), there will be merge conflicts in the `api.on(...)` registration block. Recommend merging this first and rebasing the others.

Also worth noting: switching from `void` fire-and-forget to `await` inside `gateway_start` is correct — it means startup errors in `markdownIngestion.start()` or `dreamPromotion.start()` will now propagate instead of being silently swallowed. That's an improvement, but it could change observable behavior if either service was previously failing silently on some installs.

src/index.ts
```diff
@@ -119,11 +119,13 @@ export function register(api: OpenClawPluginApi) {
     await dreamPromotion.stop();
   });
 
-  void markdownIngestion.start().catch((error) => {
-    api.logger?.warn?.(`LibraVDB markdown ingestion failed to start: ${error instanceof Error ? error.message : String(error)}`);
-  });
-  void dreamPromotion.start().catch((error) => {
-    api.logger?.warn?.(`LibraVDB dream promotion failed to start: ${error instanceof Error ? error.message : String(error)}`);
+  api.on("gateway_start", async () => {
```
fuller-stack-dev, line 122: Required: this PR is based on the pre-`registerService` startup code. Current `main` already starts/stops markdown ingestion and dream promotion through `api.registerService` with named services and non-fatal startup handling. Reintroducing a `gateway_start` hook here conflicts with that path and would lose the service-registry semantics once rebased. Please rebase onto `main` and either close this as superseded or adapt the current `registerService` implementation instead.

fuller-stack-dev (changes requested):
Requested changes: this branch is stale against `main` and the lifecycle change is now superseded by the service registry path.

xDarkicex:
fixed in PR #75 closed

PR #73, opened 2026-05-03
fix: add agent_end hook for per-turn observability

xDarkicex:
## Summary
- Registers an `agent_end` lifecycle hook that logs turn completion with success state, duration, and error details
- Observation-only — no blocking calls, no RPC, lightweight logging
- Fires per-turn (not per-session), complementing the existing `session_end` hook

## Configuration required
Non-bundled plugins must set in their OpenClaw config:
```json
{
  "plugins": {
    "entries": {
      "libravdb-memory": {
        "hooks": {
          "allowConversationAccess": true
        }
      }
    }
  }
}
```

## Test plan
- [x] Handler is fire-and-forget (no async RPC, just logger.info)
- [x] Uses structural typing on the event to avoid strict SDK type coupling

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -131,7 +131,7 @@ export function register(api: OpenClawPluginApi) {
   api.on("agent_end", async (event: unknown) => {
     const e = asRecord(event) ?? {};
     logger.info?.(
-      `LibraVDB agent_end success=${e.success ?? false} ` +
+      `LibraVDB agent_end success=${e.success ?? 'unknown'} ` +
       `durationMs=${e.durationMs ?? "?"} ` +
       `error=${typeof e.error === "string" ? e.error : "none"}`,
     );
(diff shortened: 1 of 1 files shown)


compoodment:
Minor concern: the fallback `e.success ?? false` means any `agent_end` event that omits the `success` field logs as `success=false`. If OpenClaw ever sends `agent_end` without that field (e.g., for a non-error completion), this would be misleading in logs. Consider `e.success ?? true` or logging the field only when present: `success=${e.success ?? 'unknown'}`.

Separately: this PR conflicts with #72 and #75 in `src/index.ts` around the `api.on(...)` registration block. Recommend rebasing after whichever of those merges first.

fuller-stack-dev (dismissed):
Reviewed the agent_end observability hook; no blocking issues found.

xDarkicex:
> Minor concern: the fallback `e.success ?? false` means any `agent_end` event that omits the `success` field logs as `success=false`. If OpenClaw ever sends `agent_end` without that field (e.g., for a non-error completion), this would be misleading in logs. Consider `e.success ?? true` or logging the field only when present: `success=${e.success ?? 'unknown'}`.
> 
> Separately: this PR conflicts with #72 and #75 in `src/index.ts` around the `api.on(...)` registration block. Recommend rebasing after whichever of those merges first.

Great find, Thank you @compoodment

PR #75, opened 2026-05-03
fix: add registerRuntimeLifecycle for cleanup on plugin disable

xDarkicex:
## Summary
- Registers an `onDisable` callback via `api.registerRuntimeLifecycle()` that shuts down the sidecar runtime
- Covers the disable-at-runtime path that `gateway_stop` does not fire for (plugin disabled via config while gateway is running)
- `runtime.shutdown()` is idempotent — guarded by `stopped`/`shuttingDown` flags preventing double-shutdown
- Adds `registerRuntimeLifecycle` to local SDK type declarations with optional chaining for SDK version compat

## Test plan
- [x] Optional chaining used — graceful no-op on SDK versions without this API
- [x] Double-shutdown is safe: `plugin-runtime.ts` guards against it

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -140,6 +140,13 @@ export function register(api: OpenClawPluginApi) {
     },
   });
 
+  api.registerRuntimeLifecycle?.({
+    async onDisable() {
+      logger.info?.("LibraVDB disable — shutting down runtime");
+      await runtime.shutdown();
+    },
+  });
+
   api.on("before_reset", createBeforeResetHook(runtime, api.logger ?? console));
   api.on("session_end", createSessionEndHook(runtime, api.logger ?? console));
   api.on("gateway_stop", async () => {
--- a/src/index.ts
+++ b/src/index.ts
@@ -141,9 +141,13 @@ export function register(api: OpenClawPluginApi) {
   });
 
   api.registerRuntimeLifecycle?.({
-    async onDisable() {
-      logger.info?.("LibraVDB disable — shutting down runtime");
-      await runtime.shutdown();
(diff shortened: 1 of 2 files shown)


compoodment:
Two notes:

1. The type declaration adds `onReload?`, `onDelete?`, `onReset?` callbacks, but only `onDisable` is wired. That's fine for forward compatibility, but worth a brief comment in the type or the PR description so it's clear these are placeholder declarations, not silently dropped handlers.

2. Same merge order note as #72 and #73: all three PRs modify adjacent code in `src/index.ts`. Rebasing after #72 merges would avoid conflicts.

src/index.ts
```diff
@@ -140,6 +140,13 @@ export function register(api: OpenClawPluginApi) {
     },
   });
 
+  api.registerRuntimeLifecycle?.({
```
fuller-stack-dev, line 143: Required: this does not match the current OpenClaw `registerRuntimeLifecycle` contract. In the latest SDK the registration is `{ id, description?, cleanup(ctx) }`, with `ctx.reason` values like `disable`, `reset`, `delete`, and `restart`; it does not call `onDisable`. The repo lock (`openclaw@2026.4.11`) also does not expose this API yet, so optional chaining is fine for compatibility, but hosts that do expose it will ignore this callback shape. Please register an `id` plus `cleanup` and gate on `ctx.reason === "disable"` if disable-only shutdown is intended, and update the local type declaration to the same shape.

fuller-stack-dev (changes requested):
Requested changes: the runtime lifecycle registration shape does not match the OpenClaw SDK contract.

PR #78, opened 2026-05-03
fix: apply default threshold to memory search

### LINKED ISSUE #77: memory search returns weak matches by default without applying score threshold
## Environment

- OS: WSL2 Linux x64
- OpenClaw: 2026.4.23
- Plugin: `@xdarkicex/openclaw-memory-libravdb` v1.4.35
- Daemon: `libravdbd` v1.4.29
- Embedding profile: `all-minilm-l6-v2`, ONNX local, CPU
- Test setup: isolated OpenClaw profile using `libravdb-memory` as the memory slot, connected to a local daemon over a Unix socket

## Reproduction

With at least one stored after-turn memory, run a deliberately unrelated query through the OpenClaw memory CLI:

```bash
openclaw --profile <test-profile> memory search 'ZXQ_UNRELATED_QUERY_WITH_NO_EXPECTED_MATCH_20260503'
openclaw --profile <test-profile> memory search 'ZXQ_UNRELATED_QUERY_WITH_NO_EXPECTED_MATCH_20260503' --min-score 0.35
```

For control, run a relevant query that should match the stored memory:

```bash
openclaw --profile <test-profile> memory search '<known-memory-marker>'
openclaw --profile <test-profile> memory search '<known-memory-marker>' --min-score 0.35
```

## Observed behavior

The unrelated query returns a low-scoring memory by default:

```text
0.253 <memory-result-path>
<stored memory text unrelated to the query>
```

The same unrelated query with `--min-score 0.35` correctly returns no matches:

```text
No matches.
```

The relevant query returns the expected memory both with and without `--min-score 0.35`:

```text
0.762 <memory-result-path>
<stored memory text matching the query>
```

## Expected behavior

Default CLI search should not return weak semantic matches below the plugin/daemon threshold unless the user explicitly lowers the threshold.

A reasonable default would be to apply the daemon/plugin gate threshold, currently shown by `memory status` as `0.35`, when `--min-score` is not provided. Alternatively, the CLI should document that default search is intentionally unbounded and recommend `--min-score` for normal usage.

## Scope assessment

- Search itself works: explicit `--min-score 0.35` filters the false positive while preserving the relevant match.
- This appears to be a CLI/default-threshold issue, not daemon availability or embedding failure.
- The behavior can make memory recall look noisy because unrelated queries produce low-confidence results instead of `No matches.`

compoodment:
## Summary

- apply the sidecar/status gate threshold as the default `memory search` `minScore`
- keep explicit `--min-score` overrides working, including lower thresholds
- update the checklist integration test for the existing `contracts` manifest key so the contributor gate passes on current main

Fixes #77.

## Verification

- `PATH=/tmp/corepack-shims:$PATH pnpm check`
- `PATH=/tmp/corepack-shims:$PATH npm run test:integration`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved search result filtering with consistent minimum score threshold handling. The search command now always applies a minimum score—either from the `--min-score` flag or intelligent defaults—ensuring predictable result filtering behavior.

* **Tests**
  * Added unit tests for search gating behavior.
  * Updated integration test expectations.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -421,12 +421,13 @@ async function runSearch(
       agentId: opts?.agent,
     });
     const maxResults = normalizeLimit(opts?.maxResults ?? opts?.limit);
-    const minScore = normalizeNumber(opts?.minScore);
+    const explicitMinScore = normalizeNumber(opts?.minScore);
+    const minScore = explicitMinScore ?? resolveDefaultSearchMinScore(manager.status(), cfg);
     const results = (await manager.search(
       {
         query,
         ...(maxResults ? { maxResults } : {}),
-        ...(minScore !== undefined ? { minScore } : {}),
+        minScore,
       },
     )) as Array<{
       path: string;
@@ -452,6 +453,10 @@ async function runSearch(
     logger.error(`LibraVDB search failed: ${formatError(error)}`);
     process.exitCode = 1;
   }
+}
+
+function resolveDefaultSearchMinScore(status: { gatingThreshold?: number } | undefined, cfg: PluginConfig): number {
+  return normalizeNumber(status?.gatingThreshold) ?? normalizeNumber(cfg.ingestionGateThreshold) ?? 0.35;
 }
 
 async function runFlush(runtime: PluginRuntime, opts: CliOptionBag | undefined, logger: LoggerLike): Promise<void> {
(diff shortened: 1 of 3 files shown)


compoodment:
Addressed CodeRabbit's test nit in 672b108: the default-threshold test now uses a status gatingThreshold of 0.5 and includes a 0.40 mid-range result, so it fails if the code falls back to the hardcoded 0.35 instead of reading manager.status(). Re-ran gates: `PATH=/tmp/corepack-shims:$PATH pnpm check` and `PATH=/tmp/corepack-shims:$PATH npm run test:integration` both pass.

fuller-stack-dev (approved):
Reviewed the threshold fix and updated tests; no blocking issues found.

xDarkicex:
good idea, thank you @compoodment

PR #85, opened 2026-05-06
fix: preserve runtime on plugin disable

fuller-stack-dev:
## Summary
- Avoid shutting down the LibraVDB runtime when OpenClaw runs plugin disable lifecycle cleanup during an active turn
- Keep runtime shutdown for destructive lifecycle cleanup while preserving the existing gateway_stop shutdown path
- Add unit coverage for lifecycle reason handling

## Testing
- pnpm run check
- pnpm exec tsc -p tsconfig.tests.json && node --test .ts-build/test/integration/checklist-validation.test.js
- pnpm run plugin:check

## Summary by CodeRabbit

* **New Features**
  * Enhanced runtime lifecycle management to properly handle shutdown for reset, delete, and restart operations, while preserving runtime during disable operations.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -13,6 +13,11 @@ import type { PluginConfig } from "./types.js";
 export const MEMORY_ID = "libravdb-memory";
 
 const LIGHTWEIGHT_MODES = new Set(["cli-metadata", "setup-only"]);
+type RuntimeLifecycleCleanupReason = "disable" | "reset" | "delete" | "restart";
+
+export function shouldShutdownRuntimeForLifecycleReason(reason: RuntimeLifecycleCleanupReason): boolean {
+  return reason !== "disable";
+}
 
 export function register(api: OpenClawPluginApi) {
   const registrationMode = api.registrationMode;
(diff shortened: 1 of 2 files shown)


compoodment (changes requested):
This feels like a hack that trades a crash for a permanent process leak.

PR #75 explicitly added this hook because `gateway_stop` does not fire when a plugin is disabled at runtime. If you return without calling `await runtime.shutdown()` on `disable`, the sidecar process will just sit in the background consuming RAM and holding port locks until the gateway eventually restarts. If the user re-enables the plugin later, it'll spin up a *second* vector DB process.

If the root problem is that disabling a plugin mid-turn crashes the active turn, the correct fix is likely in OpenClaw core (it should drain active tasks before firing cleanup hooks) or you need to delay the shutdown here until the turn completes. Just dropping the `disable` shutdown entirely breaks the live-unload path completely.

Also, minor nits (similar to CodeRabbit's): if you're going to define `RuntimeLifecycleCleanupReason`, export it so callers can type-check it, and interpolate `ctx.reason` in the log instead of hardcoding `disable`. But the structural leak needs to be addressed first.

xDarkicex:
Closing — this was already handled correctly in #75 by compoodment. The current behavior (shutdown on disable) is intentional: failing to shut down the runtime on plugin disable would leak the vector service process, consuming RAM and holding port locks. If the user re-enables the plugin, they'd get a second conflicting process. Thanks compoodment for catching this.

Issue #94, opened 2026-05-07
bug: gRPC HMAC challenge-response uses key and message backwards


compoodment:
## Summary

The `GrpcKernelClient.getMetadata()` method constructs the HMAC signature with the nonce as the key and the shared secret as the message. Standard challenge-response authentication signs the server-issued nonce with the shared secret:

```ts
// Current (backwards):
const hmac = createHmac("sha256", Buffer.from(this.nonceHex, "hex"));
hmac.update(this.secret);

// Expected:
const hmac = createHmac("sha256", Buffer.from(this.secret, "utf8"));
hmac.update(this.nonceHex);
```

In a challenge-response protocol, the server sends a nonce to prevent replay, and the client proves possession of the shared secret by computing `HMAC(secret, nonce)`. The current code computes `HMAC(nonce, secret)` — the key and message are swapped.

This means **no gRPC-authenticated request can succeed** against a server that validates the HMAC correctly (i.e., computes `HMAC(secret, nonce)` on its side and compares). If this has ever worked, it would only be because the server also has the same swapped construction.

**Also:** the nonce is stored as hex and used as hex-bytes (`Buffer.from(this.nonceHex, "hex")`), but the secret is treated as UTF-8 (`this.secret` is a plain string, so `hmac.update(this.secret)` encodes it as UTF-8). If the server expects both to be handled consistently (both hex-decoded or both UTF-8), this is a second inconsistency.

## Verification

The bug is in `src/grpc-client.ts`, lines 55-57:

```ts
const hmac = createHmac("sha256", Buffer.from(this.nonceHex, "hex"));
hmac.update(this.secret);
const signature = hmac.digest("hex");
```

The `initializeSession` response extracts the nonce from `resp.server_metadata.nonce` (line 84-90) and stores it as `this.nonceHex` — a hex string. The secret comes from `LIBRAVDB_AUTH_SECRET` env or `LIBRAVDB_AUTH_SECRET_FILE`.

## Impact

- Any deployment using `LIBRAVDB_AUTH_SECRET` for gRPC authentication will fail all authenticated RPCs after `initializeSession`.
- If the daemon also has the swapped construction, both sides are computing the same wrong HMAC — so it "works" but is cryptographically incorrect (the nonce is being used as the HMAC key, which defeats the purpose of a challenge-response since the nonce is sent in the clear).

## Proposed Fix

Swap the HMAC key and message:

```ts
const hmac = createHmac("sha256", this.secret);
[...]

Issue #76, opened 2026-05-03
rebuild_index fails with version conflicts on fresh v1.4.29 store


compoodment:
## Environment

- OS: WSL2 Linux x64
- OpenClaw: 2026.4.23
- Plugin: `@xdarkicex/openclaw-memory-libravdb` v1.4.32
- Daemon: `libravdbd` v1.4.29
- Embedding profile: `all-minilm-l6-v2`, ONNX local, CPU
- Test setup: isolated OpenClaw profile using `libravdb-memory` as the memory slot, connected to a local daemon over a Unix socket

## Background

After updating daemon v1.4.26 -> v1.4.29, the old store was rejected due to the expected embedding fingerprint change:

```text
embedding profile mismatch: store fingerprint <old-fingerprint> does not match current fingerprint <new-fingerprint>
```

I moved the old disposable DB aside and let v1.4.29 initialize a fresh store. Daemon then started cleanly and `openclaw memory status` under the isolated test profile reported `Sidecar: running`, `Message: ok`, `Turns stored: 0`, `Memories stored: 0`.

## Reproduction

On the fresh v1.4.29 store, using an isolated OpenClaw profile configured with `libravdb-memory`:

```bash
openclaw --profile <test-profile> memory status
openclaw --profile <test-profile> memory index --force
openclaw --profile <test-profile> memory status --deep
```

I also reproduced directly over the daemon RPC:

```js
await rpc.call('rebuild_index', { namespace: '' })
await rpc.call('rebuild_index', { namespace: '', collections: ['authored:hard'] })
await rpc.call('rebuild_index', { namespace: '', collections: ['authored:soft'] })
await rpc.call('rebuild_index', { namespace: '', collections: ['authored:variant'] })
```

## Observed behavior

CLI rebuild fails with version conflicts / missing records even though this is a freshly initialized v1.4.29 store:

```text
Collections processed: 0
Records reindexed:     0
[plugins] LibraVDB index rebuild: authored:hard: tx: transaction commit failed: record not found: <workspace-file>#000003
[plugins] LibraVDB index rebuild: authored:soft: tx: transaction commit failed: version conflict for authored:soft/<workspace-file>#000001: expected version 0, actual version 1
[plugins] LibraVDB index rebuild: authored:variant: tx: transaction commit failed: record not found: <workspace-file>#000010
[plugins] LibraVDB index rebuild completed with errors and no records reindexed.
```

[...]

PR #98, opened 2026-05-07
fix: reset sidecar retry counter on successful reconnect

### LINKED ISSUE #95: bug: sidecar retry counter never resets after successful reconnect
## Summary

The `SidecarSupervisor` class tracks daemon crash/reconnect retries in `this.retries`, but this counter is **never reset to 0 after a successful reconnect**. The counter is initialized to 0 at construction and only ever incremented in `handleExit()`:

```ts
// sidecar.ts line 231
private retries = 0;

// line 330
this.retries += 1;
```

After `start()` successfully connects, `retries` keeps its accumulated value. This means:

- If the daemon crashes 4 times over the **entire lifetime** of the gateway process (not 4 times in a row), the plugin permanently enters degraded mode and stops attempting reconnects.
- The retry exhaustion threshold (`maxRetries`, default 3) is effectively a lifetime budget, not a per-incident budget.

## Steps to Reproduce

1. Configure a LibraVDB plugin with default `maxRetries` (3).
2. Start the gateway. Daemon connects successfully.
3. Kill the daemon. Plugin reconnects (retry 1). Restart daemon, plugin connects.
4. Kill the daemon again. Plugin reconnects (retry 2). Restart daemon, plugin connects.
5. Kill the daemon a third time. Plugin reconnects (retry 3). Restart daemon, plugin connects.
6. Kill the daemon a fourth time. Plugin enters **degraded mode permanently** — no further reconnection attempts.

Expected: each successful reconnect should reset the counter. The daemon should tolerate 3 consecutive failures, not 3 cumulative failures over the entire process lifetime.

## Proposed Fix

Reset `this.retries` to 0 in `start()` after a successful connection:

```ts
async start(): Promise<SidecarSocket> {
  const endpoint = await this.runtime.resolveEndpoint(this.cfg);
  const socket = await this.connectEndpointWithRetry(endpoint);
  this.retries = 0;  // Reset on successful connection
  // ...
}
```

compoodment:
## Summary

The `SidecarSupervisor` retry counter is never reset after a successful reconnect. The counter accumulates over the entire gateway lifetime, so the plugin permanently enters degraded mode after `maxRetries` (default 3) **cumulative** daemon crashes — not 3 consecutive failures.

- **Bug:** `this.retries` starts at 0, is incremented in `handleExit()`, and is never reset. After 3+ daemon crashes over the lifetime of the gateway process, the plugin gives up permanently.
- **Fix:** Add `this.retries = 0` in `start()` after a successful connection, so the retry budget resets per-incident rather than being lifetime-cumulative.

Closes #95

## Verification

Before: kill and restart the daemon 4 times over a gateway lifetime → permanent degraded mode after the 3rd crash, even though each crash was followed by a successful reconnect.

After: the retry counter resets to 0 on each successful reconnect, so the plugin tolerates up to 3 consecutive failures per incident.

> **Daemon source note:** This is a plugin-only fix, but we continue to note that issues like #76 (rebuild_index version conflicts) and #94 (gRPC HMAC key/message swap) require access to the daemon source code for a complete resolution. We are unable to submit fixes for daemon-side bugs without it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of internal service recovery handling by ensuring the retry counter resets properly during startup.

– Vale

### DIFF
--- a/src/sidecar.ts
+++ b/src/sidecar.ts
@@ -245,6 +245,7 @@ class SidecarSupervisor implements SidecarHandle {
   async start(): Promise<SidecarSocket> {
     const endpoint = await this.runtime.resolveEndpoint(this.cfg);
     const socket = await this.connectEndpointWithRetry(endpoint);
+    this.retries = 0;
     this.reconnectScheduled = false;
     if (this.socket instanceof SupervisorSocket) {
       this.socket.bind(socket);


xDarkicex:
Safe to merge, targeted fix thank you @compoodment

PR #105, opened 2026-05-07
fix: account for system prompt additions in assemble budget clamp

### LINKED ISSUE #104: bug: context-engine budget clamp ignores systemPromptAddition
## Summary

`enforceTokenBudgetInvariant()` only considers replay messages when clamping assemble output to `tokenBudget`. It ignores `systemPromptAddition`, even though that text is injected into the final prompt and consumes budget.

Current logic:

```ts
const estimated = typeof result.estimatedTokens === "number" ? result.estimatedTokens : 0;
const approxFromMessages = approximateMessagesTokens(result.messages);

if (estimated <= effectiveBudget && approxFromMessages <= effectiveBudget) {
  return result;
}

const trimmedMessages = trimMessagesToBudget(result.messages, effectiveBudget);
```

If the daemon/kernel returns a huge `systemPromptAddition` with a low or stale `estimatedTokens`, the invariant passes and the full addition is returned unchanged.

## Reproduction

Using the current public code:

```ts
assemble_context_internal -> {
  messages: [],
  estimatedTokens: 0,
  systemPromptAddition: "x".repeat(2000),
}

tokenBudget: 100
```

Actual returned result:

```txt
estimatedTokens: 0
systemPromptAddition.length: 2000
messages.length: 0
```

Expected: the budget clamp should account for `systemPromptAddition` and either reduce available replay-message budget or clamp/truncate/drop the addition.

## Impact

The context engine fallback/budget guard can still exceed host prompt budgets when the sidecar/kernel emits large system prompt additions or underestimates `estimatedTokens`.

This is especially relevant because exact recall already treats system additions as budget-consuming text (`additionTokens`). The final invariant should do the same for any daemon/kernel `systemPromptAddition`.

## Proposed Fix

Include `approximateTokenCount(result.systemPromptAddition)` in the budget invariant:

- return unchanged only when `estimatedTokens` and `systemPromptAddition + messages` fit,
- reserve budget for `systemPromptAddition` before trimming replay messages,
- if the addition alone exceeds the effective budget, clamp the addition and drop replay messages.

compoodment:
## Summary

`enforceTokenBudgetInvariant()` only considered replay messages when clamping context-engine assemble output to `tokenBudget`. It ignored `systemPromptAddition`, even though that text is also injected into the final prompt and consumes budget.

If the daemon/kernel returned a large `systemPromptAddition` with stale or underreported `estimatedTokens`, the invariant could pass and return the full addition unchanged.

This PR:

- includes `systemPromptAddition` in the approximate budget calculation,
- returns unchanged only when both `estimatedTokens` and approximate prompt content fit,
- reserves budget for `systemPromptAddition` before trimming replay messages,
- truncates the system prompt addition and drops replay messages when the addition alone exceeds the effective budget,
- drops replay messages when the residual budget cannot even cover message wrapper overhead,
- uses head-preserving truncation for system prompt additions, so XML-ish/preamble structure is not preferentially discarded,
- adds unit coverage for oversized additions, additions that fit while forcing message trimming, and near-zero residual message budgets.

Closes #104.

## Verification

- `pnpm exec tsc -p tsconfig.tests.json && node --test .ts-build/test/unit/context-engine.test.js` — PASS
- `pnpm run check` — PASS, 110 tests

Follow-up commit `5ce9860` fixed the wrapper-overhead residual-budget edge case found during audit.

– Vale

### DIFF
--- a/src/context-engine.ts
+++ b/src/context-engine.ts
@@ -330,7 +330,10 @@ function trimMessagesToBudget(
   }
 
   const last = messages[messages.length - 1]!;
-  const contentBudget = Math.max(1, tokenBudget - 8);
+  const contentBudget = tokenBudget - 8;
+  if (contentBudget <= 0) {
+    return [];
+  }
   const truncated = truncateContentToTokenBudget(last.content, contentBudget);
   if (!truncated) {
     return [];
@@ -371,7 +374,7 @@ function enforceTokenBudgetInvariant(
     };
   }
 
-  const messageBudget = Math.max(1, effectiveBudget - systemPromptTokens);
+  const messageBudget = Math.max(0, effectiveBudget - systemPromptTokens);
   const trimmedMessages = trimMessagesToBudget(result.messages, messageBudget);
   const trimmedEstimate = approximateMessagesTokens(trimmedMessages);
   return {
--- a/src/context-engine.ts
+++ b/src/context-engine.ts
@@ -297,6 +297,14 @@ function truncateContentToTokenBudget(content: unknown, tokenBudget: number): st
   return normalized.slice(normalized.length - maxChars);
 }
 
+function truncateSystemPromptAdditionToTokenBudget(value: string, tokenBudget: number): string {
+  if (tokenBudget <= 0) return "";
+  const maxChars = Math.max(1, tokenBudget * APPROX_CHARS_PER_TOKEN);
+  if (value.length <= maxChars) return value;
+  // System additions are head-structured: preserve XML/preamble/instructions.
+  return value.slice(0, maxChars);
+}
+
 function trimMessagesToBudget(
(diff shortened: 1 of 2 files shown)


compoodment:
Addressed the CodeRabbit finding in `6b82afe`: system prompt additions now use a head-preserving truncator instead of the tail-preserving message-content truncator, and the unit test asserts the prefix is preserved. Re-ran targeted context-engine unit test and full `pnpm run check` (107 tests pass).

– Vale

compoodment:
Addressed the remaining CodeRabbit coverage nit in `fix/context-budget-system-prompt`: added a unit test for the branch where `systemPromptAddition` fits but replay messages are trimmed against the remaining budget. Validation: targeted context-engine unit test passes (30 tests) and full `pnpm run check` passes (108 tests).

– Vale

PR #107, opened 2026-05-08
fix: replace first chunk during markdown ingest

### LINKED ISSUE #106: bug: chunked markdown ingest replaces earlier chunks with the final chunk
## Summary

The new chunked markdown ingest path sends multi-chunk documents with `APPEND` for every chunk except the final chunk, then sends the final chunk with `REPLACE`.

`IngestMode.REPLACE` is documented in the generated contracts as “delete existing nodes, then insert new”, while `APPEND` means “skip delete, only insert new”. That ordering means a multi-chunk document can ingest chunks 1..N-1, then the final `REPLACE` call deletes those earlier chunks and leaves only chunk N indexed.

## Repro shape

For any markdown document larger than the chunk threshold:

1. `splitIntoChunks()` returns multiple chunks.
2. `enqueueIngest()` sends:
   - chunk 0: `APPEND`
   - chunk 1: `APPEND`
   - final chunk: `REPLACE`
3. The daemon receives the final call as a replacement for the same `sourceDoc`, so prior appended nodes for that source are deleted.

## Expected

Replacement should happen before appends for the same document, e.g. first chunk `REPLACE`, later chunks `APPEND`, so the full document remains indexed.

## Impact

Large markdown/core documents silently lose all but their final chunk. Search/recall will look like ingestion succeeded, but most of the document content is gone from the authored collection.

## Verification

Checked current `src/ingest-queue.ts` and generated `IngestMode` contract comments in `@xdarkicex/libravdb-contracts`.

compoodment:
## Summary
- send the first chunk of a multi-chunk markdown ingest as REPLACE
- send remaining chunks as APPEND so earlier chunks are not deleted by the final call
- add unit coverage for chunk ingest mode ordering and content preservation

## Verification
- pnpm exec tsc -p tsconfig.tests.json
- node --test .ts-build/test/unit/ingest-queue.test.js
- pnpm run check

Fixes #106

### DIFF
--- a/src/ingest-queue.ts
+++ b/src/ingest-queue.ts
@@ -83,14 +83,14 @@ export class IngestQueue {
       });
     }
 
-    // Multiple chunks: use APPEND mode for all but last (which can be REPLACE)
+    // Multiple chunks: replace the previous document first, then append remaining chunks.
     for (let i = 0; i < chunks.length; i++) {
-      const isLast = i === chunks.length - 1;
+      const isFirst = i === 0;
       const chunkParams: IngestMarkdownDocumentParams = {
         ...baseParams,
         sourceDoc,
         text: chunks[i].text,
-        mode: isLast ? IngestMode.REPLACE : IngestMode.APPEND,
+        mode: isFirst ? IngestMode.REPLACE : IngestMode.APPEND,
       };
       await this.ingestWithRetry(chunkParams);
     }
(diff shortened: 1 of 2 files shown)


compoodment:
Rebased this fix onto upstream `v1.4.47` (`e904b7a`) after the forced main/tag update. Validation on the rebased branch: `pnpm run check` passes (107 tests).

compoodment:
This fix is already included in v1.4.48 (shipped on the main branch). Closing as resolved.

PR #112, opened 2026-05-08
fix: swap HMAC key and message in gRPC challenge-response

### LINKED ISSUE #94: bug: gRPC HMAC challenge-response uses key and message backwards

compoodment:
## Summary

The gRPC client computed the auth signature as `HMAC(nonce, secret)`: the server nonce was used as the HMAC key and the shared secret was used as the message.

Challenge-response auth should prove possession of the shared secret by signing the server-issued nonce:

```ts
HMAC(secret, nonce)
```

This PR changes the client to use the configured shared secret as the HMAC key and the hex-encoded nonce string as the message, matching the protocol expectation documented here.

## Compatibility warning

This is a coordinated protocol fix. If the daemon currently validates the same swapped construction, a client-only change will break authenticated gRPC calls until the daemon validates `HMAC(secret, nonce)` too.

So: correct fix, but do not merge/deploy blindly without daemon-side alignment.

Closes #94.

## Verification

- `pnpm exec tsc -p tsconfig.tests.json && node --test .ts-build/test/unit/grpc-client.test.js` — PASS
- `pnpm run check` — PASS, 119 tests

The new tests use a fixed HMAC vector and assert that signed metadata contains the expected `x-libravdb-auth` signature. They also check that signed calls still require an initialized server nonce. Follow-up merge commit `2f376fa` resolved the current-main `grpc-client.test.ts` conflict while keeping both the target-resolution and HMAC coverage.

– Vale

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed authentication credential generation in gRPC client to ensure secure and proper communication between client and server

* **Tests**
  * Expanded test suite to verify authentication flows, including session initialization requirements and credential validation

[Review Change Stack](https://app.coderabbit.ai/change-stack/xDarkicex/openclaw-memory-libravdb/pull/112)

### DIFF
--- a/src/grpc-client.ts
+++ b/src/grpc-client.ts
@@ -52,8 +52,12 @@ export class GrpcKernelClient {
       if (!this.nonceHex) {
         throw new Error("call initializeSession before authenticated RPCs");
       }
-      const hmac = createHmac("sha256", Buffer.from(this.nonceHex, "hex"));
-      hmac.update(this.secret);
+      // Challenge-response: HMAC(secret, nonce) — the secret is the HMAC key,
+      // the server-issued nonce is the message. The previous implementation
+      // swapped these, computing HMAC(nonce, secret), which is cryptographically
+      // incorrect: the nonce is sent in the clear and must not be used as the key.
+      const hmac = createHmac("sha256", this.secret);
+      hmac.update(this.nonceHex);
       const signature = hmac.digest("hex");
       md.add("x-libravdb-auth", signature);
     }
--- a/test/unit/grpc-client.test.ts
+++ b/test/unit/grpc-client.test.ts
@@ -1,0 +1,36 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+
+import { GrpcKernelClient } from "../../src/grpc-client.js";
+
+type MetadataProbe = {
+  secret?: string;
+  nonceHex?: string;
+  getMetadata(signed?: boolean): { get(key: string): unknown[] };
+};
+
+function createMetadataProbe(secret?: string, nonceHex?: string): MetadataProbe {
+  return Object.assign(
+    Object.create(GrpcKernelClient.prototype),
+    { secret, nonceHex },
+  ) as MetadataProbe;
+}
+
+test("gRPC auth metadata signs the server nonce with the shared secret", () => {
+  const client = createMetadataProbe("test-secret", "00112233445566778899aabbccddeeff");
+
+  const metadata = client.getMetadata(true);
+
+  assert.deepEqual(metadata.get("x-libravdb-auth"), [
+    "84b7660ccd62a3d5848a05112cd1cff4e753779cb4464c156e0e57d7c9b6cef3",
+  ]);
+});
+
+test("gRPC auth metadata requires initializeSession nonce before signed calls", () => {
+  const client = createMetadataProbe("test-secret");
+
+  assert.throws(
+    () => client.getMetadata(true),
+    /call initializeSession before authenticated RPCs/,
+  );
+});


compoodment:
**Note on daemon dependency:** This PR fixes the plugin side of the HMAC swap, but the daemon (`libravdbd`) also needs a corresponding change to validate `HMAC(secret, nonce)` instead of the current `HMAC(nonce, secret)`. Until the daemon is updated, deploying this PR alone will break gRPC authentication for any deployment using `LIBRAVDB_AUTH_SECRET`.

If the daemon currently computes the same swapped HMAC, both sides produce the same (wrong) result — so the existing "works but is cryptographically incorrect" state continues until both are updated simultaneously.

Possible deployment strategies:
1. **Coordinate release** — ship plugin + daemon updates together, both switch to `HMAC(secret, nonce)` at the same time.
2. **Dual-validation** — have the daemon accept both `HMAC(secret, nonce)` and `HMAC(nonce, secret)`, then remove the old form in a later version.

– Vale

compoodment:
Cross-PR blocker: this PR and #145 are individually mergeable, but they conflict as soon as both are stacked.

Repro on current `origin/main`:

```bash
git merge --no-edit pr-112
git merge --no-edit pr-145
```

Actual conflict:

```text
CONFLICT (content): Merge conflict in test/unit/grpc-client.test.ts
```

The conflict is just the shared grpc unit test file: #112 adds `GrpcKernelClient` auth metadata coverage, while #145 adds `resolveGrpcCredentialMode` TLS coverage. Reverse order fails the same way.

Please rebase after whichever grpc PR lands first, or pre-combine the tests so both grpc fixes can merge without creating an immediate follow-up conflict.

PR #114, opened 2026-05-08
fix(markdown-ingest): detect CRLF frontmatter in looksLikeObsidianNote

compoodment:
## Summary

`looksLikeObsidianNote` gated frontmatter detection with `text.startsWith("---\n")`, which fails for CRLF files starting with `"---\r\n"`. Those files were sent directly to `hasInlineObsidianTag` without ever parsing frontmatter for `tags:`, `memory:`, or `openclaw:` keys.

PR #101 fixed `findFrontmatterEnd` to handle CRLF closing delimiters, but missed the upstream gate that decides *whether* to call it at all.

## Fix

Accept both `"---\n"` and `"---\r\n"` as frontmatter openers. Pass the correct offset (4 for LF, 5 for CRLF) to `findFrontmatterEnd`.

## Verification

- TypeScript compiles clean
- A CRLF file with `---\r\ntags: [memory]\r\n---\r\n` is now correctly detected as an Obsidian note

– Vale

### DIFF
--- a/src/markdown-ingest.ts
+++ b/src/markdown-ingest.ts
@@ -595,17 +595,20 @@ function matchesGlob(value: string, pattern: string): boolean {
 }
 
 function looksLikeObsidianNote(filePath: string, text: string): boolean {
-  if (!text.startsWith("---\n")) {
+  const isCrlf = text.startsWith("---\r\n");
+  const hasFrontmatterOpen = isCrlf || text.startsWith("---\n");
+  if (!hasFrontmatterOpen) {
     return hasInlineObsidianTag(text);
   }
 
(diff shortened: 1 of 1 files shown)


compoodment:
Closing after re-audit: #101 covers the real user-facing cases (CRLF frontmatter + heading tags) with tests. This PR is the narrower duplicate and has no tests, so keeping both open just creates review noise.\n\nNote: the closing-delimiter offset difference here is not enough to justify a separate PR; with CRLF it leaves the linefeed as an empty first body line, which does not break inline tag scanning. If we want exact slicing later, fold it into #101 or a tiny cleanup, not a competing PR.\n\n– Vale

PR #116, opened 2026-05-08
fix(memory-runtime): add defensive guards for undefined metadata and results

compoodment:
## Summary

Two defensive guards in `memory-runtime.ts`:

1. **`toMemorySearchResult`**: `item.metadata` could be undefined if the daemon returns a result without the metadata field (or via a path that bypasses the protobuf codec's `normalizeSearchTextResponse`). Now destructures with `?? {}` fallback. Also guards `item.text` (default `""`) and `item.score` (default `0`) against undefined values.

2. **`loadSearchResultText`**: `result.results` could be undefined or non-array if the daemon returns a malformed response. Now normalizes to `[]` before `.find()`.

These are defensive — the protobuf codec layer already normalizes most of these — but they protect against raw gRPC or test paths that bypass the codec, and align with the same pattern used in PR #110 for the exact-recall path.

## Verification

- TypeScript compiles clean
- No behavioral change for well-formed responses
- Crashes prevented for malformed/undefined fields

– Vale

### DIFF
--- a/src/memory-runtime.ts
+++ b/src/memory-runtime.ts
@@ -229,13 +229,16 @@ function firstString(...values: Array<string | undefined>): string | undefined {
 }
 
 function toMemorySearchResult(item: SearchResult) {
-  const collection = typeof item.metadata.collection === "string" ? item.metadata.collection : "memory";
+  const metadata = item.metadata ?? {};
+  const collection = typeof metadata.collection === "string" ? metadata.collection : "memory";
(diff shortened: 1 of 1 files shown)


compoodment:
Closing after re-audit: this defensive-guard PR is superseded by the normalization landed in #110. The search/list RPC paths feeding these consumers already go through `normalizeSearchTextResponse`, so the stronger fix is at the codec boundary, not scattered consumption-site guards.\n\n– Vale

PR #119, opened 2026-05-10
fix: prevent TypeError when search results have missing text field in exact recall

JuanHuaXu:
The Bug:
In isExactRecallFact, text is expected to be a string. If LibraVDB's search returns a result where the text field is missing or null, text.includes(token) will throw the TypeError we saw.

Similarly, in rankExactRecallCandidate, result.text is accessed without a null check before calling .includes(token).

Why it triggers with Skillporter:
When you search for "security researcher," the Exact Recall logic kicks in. It tries to find facts in your memory that "mean" security researcher. If it finds a broken or empty entry in the database (maybe an old failed turn), result.text is undefined, and it crashes.

The Fix:
I need to add defensive null checks to isExactRecallFact and rankExactRecallCandidate.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved defensive input handling in recall logic to properly manage null or undefined text values.
  * Enhanced candidate ranking logic to normalize text inputs before evaluation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/xDarkicex/openclaw-memory-libravdb/pull/119)

### DIFF
--- a/src/context-engine.ts
+++ b/src/context-engine.ts
@@ -426,7 +426,8 @@ function extractExactRecallTokens(text: string): string[] {
   return Array.from(tokens).slice(0, EXACT_RECALL_MAX_TOKENS);
 }
 
-function isExactRecallFact(text: string, token: string): boolean {
+function isExactRecallFact(text: string | undefined | null, token: string): boolean {
+  if (typeof text !== "string") return false;
   return (
     text.includes(token) &&
     /\bmeans\b/i.test(text) &&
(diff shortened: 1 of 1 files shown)


compoodment (commented):
## Redundant with merged #110

This PR adds `typeof` guards at the consumption sites (`isExactRecallFact` and `rankExactRecallCandidate`), but **#110 (commit `52bd1d0`) already solved this at the codec boundary** in `normalizeSearchTextResponse`.

The codec now guarantees:

- `SearchResult.text` is always a string (empty string `""` if missing/undefined)
- `SearchResult.score` is always a finite number (`0` if missing/NaN)

By the time `isExactRecallFact` or `rankExactRecallCandidate` see the data, `text` cannot be `undefined`. The guards added here will never fire through the normal RPC path.

**The right layer for this protection is the codec boundary, not individual consumption sites** — which is exactly what #110 already does. Adding redundant guards at every consumer just masks the real guarantee and makes the codebase harder to reason about.

**Recommendation: Close this PR.** The bug it addresses is already fixed more thoroughly by #110.

---

The security findings in issues #120–#125 are separate and legitimate — those deserve their own targeted PRs.

compoodment:
## Analysis

This PR adds defensive `typeof` guards for `text` and `score` in `isExactRecallFact` and `rankExactRecallCandidate`. These are structurally redundant because PR #110 (commit `52bd1d0`) already normalizes `text` to `""` and `score` to `0` at the codec boundary in `normalizeSearchTextResponse`. Under normal RPC paths, `isExactRecallFact` and `rankExactRecallCandidate` can never receive `undefined`/`null` text.

**That said**, defense-in-depth is a legitimate posture. If the daemon ever returns a malformed response, or if a future code path bypasses the codec normalization, these guards would catch it gracefully instead of crashing.

**My recommendation:** Close as redundant (#110 already covers this), or merge as defense-in-depth if the maintainer prefers the extra safety net. Either way, this is not a bug fix — it's a hardening pass over already-protected code.

– Vale

xDarkicex:
Thanks for catching this and submitting a fix. This exact crash was addressed in #110 with a different architectural approach: instead of adding null guards at every consumption site, we normalize search results at the protobuf codec boundary (normalizeSearchTextResponse). Every search result entering the plugin has text clamped to "" and score clamped to 0 — so internal functions like isExactRecallFact and rankExactRecallCandidate can trust their input types.

The daemon side also added corresponding normalization (store.NormalizeResultText, store.NormalizeResultScore) so the raw results are safe before serialization.

Closing as already fixed by #110. Appreciate your contribution.

PR #127, opened 2026-05-11
fix: skip common irrelevant directories during markdown ingestion walk

### LINKED ISSUE #118: usability: markdown ingestion root walks generated/dependency dirs by default
## Summary

When `markdownIngestionRoots` points at a workspace directory, markdown ingestion recursively walks all `.md` files unless the user manually configures excludes. That includes dependency, VCS, cache, virtualenv, and build-output directories.

## Impact

- irrelevant README/CHANGELOG files can be ingested;
- startup scans can become slow or timeout-prone;
- retry behavior can amplify daemon load when ingestion is already busy.

## Scope

Default-exclude/usability issue. This does not solve arbitrary broad-root policy; #121 remains separate.

## Fix Path

PR #131 adds default excludes and skips excluded directories before attaching watchers.

## Dependencies

PR #131 depends on #111 because its default excludes use recursive glob patterns.

## Verification Target

- Markdown-ingest integration test proving dependency/generated dirs are skipped by default.
- Rebase #131 and retest after #111 lands.

– Vale

compoodment:
## Summary

When `markdownIngestionRoots` is set to a workspace directory, the markdown-ingest filesystem walker recurses into `node_modules/`, `.git/`, `dist/`, `build/`, `__pycache__/`, `.venv/`, `venv/`, `.next/`, `.nuxt/`, `.cache/`, `.svn/`, and `.hg/` directories. This causes hundreds of irrelevant README/CHANGELOG files to be ingested, RPC timeout cascades during startup, and bloated collections.

Add a `DEFAULT_SKIP_DIR_NAMES` set and check it before recursing into any subdirectory in `walkDirectory`. These directories are always skipped regardless of `markdownIngestionExclude` config — they never contain content that should be ingested for memory purposes.

The existing `markdownIngestionExclude` config still applies to file-level filtering. This change adds directory-level pruning that the config cannot express (previously `walkDirectory` recursed into everything).

Closes #118.

## Verification

- `tsc --noEmit` passes (pre-existing `IngestMode` import error in `ingest-queue.ts` is unrelated)
- Source-level test verifies `DEFAULT_SKIP_DIR_NAMES` contains all expected directory names
- Existing `markdown-ingest.test.ts` tests still pass

– Vale

### DIFF
--- a/src/markdown-ingest.ts
+++ b/src/markdown-ingest.ts
@@ -11,6 +11,22 @@ const DEFAULT_DEBOUNCE_MS = 150;
 const DEFAULT_TOKENIZER_ID = "markdown-ingest:v1";
 const MARKDOWN_INGEST_VERSION = 3;
 const HASH_BACKEND = "wasm-fnv1a64";
+
+/** Directories that are always skipped during markdown ingestion walks. */
+const DEFAULT_SKIP_DIR_NAMES = new Set([
+  "node_modules",
+  ".git",
+  ".svn",
+  ".hg",
+  "dist",
+  "build",
+  "__pycache__",
+  ".venv",
+  "venv",
+  ".next",
+  ".nuxt",
+  ".cache",
+]);
 type Disposable = { close(): void };
 
 interface RpcLike {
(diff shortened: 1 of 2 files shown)


compoodment:
Closing as superseded by #131, which implements the same #118 fix with nested directory coverage, docs/schema updates, and passing current verification.\n\n– Vale

PR #129, opened 2026-05-11
fix: validate collection namespace names

### LINKED ISSUE #124: 🟠 MEDIUM: Unvalidated userId allows collection name injection
### 🟠 MEDIUM: Unvalidated userId allows collection name injection

**Severity:** MEDIUM (CVSS ~5.3)  
**Affected file:** `src/memory-scopes.ts` lines 8-23  

### Summary

`resolveDurableNamespace` accepts a `userId` string and returns it as-is to the caller. This namespace is then used as a collection name in RPC calls. **No validation** of characters, length, or format.

### Taint Path

```
CLI: --user-id <userId>
  → resolveDurableNamespace()
    → returns userId as-is
      → rpc.call("search_text", { collection: userId })
```

### Impact

- Special characters may confuse daemon collection routing
- Could bypass authorization boundaries between user namespaces
- Could corrupt collection indexes

### Recommended Fix

```typescript
const COLLECTION_NAME_RE = /^[a-zA-Z][a-zA-Z0-9_:.@-]{0,127}$/;
function validateCollectionName(name: string): string {
  if (!COLLECTION_NAME_RE.test(name)) {
    throw new Error(`Invalid collection name: must match ${COLLECTION_NAME_RE}`);
  }
  return name;
}
```

compoodment:
## Summary

`resolveDurableNamespace` accepted arbitrary `userId` / fallback namespace strings and passed them through as RPC collection names. That can create confusing or unsafe namespace names (`../x`, embedded control characters, spaces, null bytes) and push validation ambiguity downstream into the daemon.

This is mostly hygiene/misconfiguration hardening, not a strong untrusted-input exploit path: the values are operator/config/session-derived. Still, validating collection names at the plugin boundary gives clearer failures and avoids malformed collection identifiers.

This PR adds `validateNamespace()` with:

```ts
/^[a-zA-Z][a-zA-Z0-9_.:@#-]{0,127}$/
```

That allows common identifiers (`user-1`, `session-key:abc`, `agent-id:my-agent`) while rejecting traversal-ish names, whitespace, null bytes, and other awkward characters.

`resolveDurableNamespace` now validates every returned namespace.

Closes #124.

## Compatibility note

This is a behavior change: fallback namespace values that previously worked despite spaces, leading digits, or other disallowed characters will now throw a descriptive error. That is intentional, but it should be treated as a config validation tightening rather than invisible behavior preservation.

## Verification

- Namespace validation unit tests cover accepted and rejected names.
- Existing `resolveDurableNamespace` tests cover the integration path.

– Vale

Closes #124

### DIFF
--- a/src/memory-scopes.ts
+++ b/src/memory-scopes.ts
@@ -1,6 +1,21 @@
 const SESSION_KEY_NAMESPACE_PREFIX = "session-key:";
 const AGENT_ID_NAMESPACE_PREFIX = "agent-id:";
 
+/** Valid collection names: alphanumeric, underscores, hyphens, dots, colons, at-signs.
+ *  Must start with a letter. Max 128 characters. */
+const COLLECTION_NAME_RE = /^[a-zA-Z][a-zA-Z0-9_.:@-]{0,127}$/;
+
+/** Validate and return a collection-safe namespace.
+ *  Throws on invalid characters or length. */
+export function validateNamespace(name: string): string {
+  if (!COLLECTION_NAME_RE.test(name)) {
+    throw new Error(
+      `Invalid collection namespace: "${name}". Must match ${COLLECTION_NAME_RE.source}`,
+    );
+  }
+  return name;
+}
+
 export type RetrievalScopes = {
   /** Always queried — fresh context bound to this session. */
   session: string;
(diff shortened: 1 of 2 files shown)


compoodment:
Found a blocking regression in this PR: the new collection-name validator rejects the plugin's own auto-derived fallback user namespace.

Repro on the PR branch:

```bash
node --test .ts-build/test/unit/memory-scopes.test.js .ts-build/test/unit/memory-runtime.test.js
```

Observed: `memory-scopes.test.js` passes, but `memory-runtime.test.js` fails in three bridge tests before RPC because `resolveDurableNamespace()` throws `Invalid collection namespace` for the auto-derived user fallback. That fallback currently includes the plugin's generated user separator, but `COLLECTION_NAME_RE` only allows `[a-zA-Z0-9_.:@-]` after the first character.

So this fix hardens explicit/config namespaces but breaks the default no-userId path. The fix probably needs either (a) allow the generated separator in the collection namespace contract, or (b) normalize/encode auto-derived identities before validation. The test coverage should include the no-explicit-user path through `buildMemoryRuntimeBridge`, not only direct `validateNamespace()` calls.

– Vale

compoodment:
Found a blocking regression in this PR: the new collection-name validator rejects the plugin's own auto-derived fallback user namespace. `resolveDurableNamespace()` throws `Invalid collection namespace` for the auto-derived user fallback, because `COLLECTION_NAME_RE` only allows `[a-zA-Z0-9_.:@-]` after the first character but auto-derived durable identities use the `user@host#hash` shape.

Pushed `2ea9342` to allow `#` in collection namespace names and added a regression assertion for an auto identity namespace.

Validated with:
```bash
node --test .ts-build/test/unit/memory-scopes.test.js
```

— Vale

Issue #121, opened 2026-05-11
🟡 LOW-MED: Markdown ingestion roots traverse beyond intended scope


JuanHuaXu:
### 🟡 LOW-MED: Markdown ingestion roots lack parent boundary check

**Severity:** LOW-MED (CVSS ~4.3)  
**Affected file:** `src/markdown-ingest.ts` lines 257, `normalizeMarkdownRoots`  

### Summary

`normalizeMarkdownRoots` resolves paths via `path.resolve()` with **no validation** that the resolved path is within an allowed parent directory.

### Taint Path

```
Config: markdownIngestionRoots (admin-set, mutable)
  → path.resolve()
    → walkDirectory(root, ...)
      → readdir/readFile on entire subtree (no bounds check)
```

### Impact

- All markdown files in the target root are indexed and searchable
- Config files like MEMORY.md, SOUL.md, IDENTITY.md may be indexed
- If root is too broad (e.g., "/"), all system markdown files are indexed

### Recommended Fix

```typescript
function normalizeMarkdownRoots(roots?: string[]): string[] {
  if (!roots?.length) return [];
  const home = os.homedir();
  const stateDir = process.env.OPENCLAW_STATE_DIR || "";
  const allowed = [home, stateDir].filter(Boolean);
  const resolved = new Set<string>();
  for (const root of roots) {
    const trimmed = root.trim();
    if (!trimmed) continue;
    const resolvedPath = path.resolve(trimmed);
    const isAllowed = allowed.some(a => resolvedPath === a || resolvedPath.startsWith(a + path.sep));
    if (!isAllowed) {
      throw new Error(`markdown ingestion root must be within home or state dir: ${resolvedPath}`);
    }
    resolved.add(resolvedPath);
  }
  return [...resolved];
}
```

compoodment:
## Assessment — valid admin footgun, not a security boundary

The concern is real: `markdownIngestionRoots` accepts operator-configured paths and will recursively walk the resolved roots. A broad root like `/` can index far more markdown than intended.

The original proposed fix is too restrictive, though. Hard-limiting roots to `$HOME` or the OpenClaw state dir would break legitimate deployments such as shared knowledge mounts, container volumes, or `/opt/knowledge`-style installs.

Better fix path:

1. Default-exclude generated/dependency/VCS directories (`node_modules`, `.git`, `dist`, `build`, `.venv`, etc.). That is covered by #118 / PR #131.
[...]

PR #131, opened 2026-05-11
fix: default-exclude generated markdown dirs

### LINKED ISSUE #118: usability: markdown ingestion root walks generated/dependency dirs by default

compoodment:
## Summary

Adds default markdown ingestion excludes for dependency, VCS, cache, virtualenv, and build-output directories.

## Scope

- Adds default exclude patterns to plugin config/schema/docs.
- Skips excluded directories during traversal before attaching watchers.
- Adds markdown-ingest integration coverage for default excludes.

## Dependencies

- Must merge after #111. This PR depends on recursive glob patterns such as `**/node_modules/**` working correctly.
- Current merge state is `DIRTY`; rebase is required before review.

## Audit Notes

- This reduces the practical workspace-footgun from generated/dependency directories.
- It does not add a hard parent-boundary policy for arbitrary configured roots; #121 remains separate.
- After rebase, verify the default excludes still match the glob semantics introduced by #111.

## Review Follow-up

- Branch is `DIRTY`; rebase after #111 before review.
- Decide whether default excludes are intentionally non-removable. If users need to ingest `.cache` or similar dirs, add an opt-out/negation design separately.
- Manifest defaults duplicate the runtime default list; acceptable for now, but keep both in sync.
- Root-level and `**/` pattern pairs are partly redundant but harmless; simplify only if tests still cover top-level and nested dirs.
- A direct `shouldSkipDirectory` unit test would be useful, though current integration coverage proves the main behavior.

## Verification

- `pnpm run check`
- `pnpm exec tsc -p tsconfig.tests.json && node --test .ts-build/test/integration/markdown-ingest.test.js`
- `npm run test:integration` was previously blocked by an unrelated degraded-sidecar test failure; markdown-ingest cases passed.

Closes #118.
Depends on #111.

– Vale

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -95,7 +95,7 @@ The plugin exposes `ingestionGateThreshold` for host-side gating decisions:
 | `markdownIngestionObsidianEnabled` | boolean | `false` | Watch Obsidian vault roots |
 | `markdownIngestionObsidianRoots` | string[] | — | Obsidian vault directories |
 | `markdownIngestionObsidianInclude` | string[] | — | Obsidian glob include patterns |
-| `markdownIngestionObsidianExclude` | string[] | — | Obsidian glob exclude patterns |
+| `markdownIngestionObsidianExclude` | string[] | common dependency/build dirs | Obsidian glob exclude patterns; merged with the same default excludes as generic markdown ingestion |
 | `markdownIngestionObsidianDebounceMs` | number | `150` | Obsidian debounce window |
 
 ## Dream promotion
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -90,7 +90,7 @@ The plugin exposes `ingestionGateThreshold` for host-side gating decisions:
 | `markdownIngestionEnabled` | boolean | `false` | Watch markdown roots for changes |
 | `markdownIngestionRoots` | string[] | — | Directories to watch |
 | `markdownIngestionInclude` | string[] | — | Glob patterns to include |
-| `markdownIngestionExclude` | string[] | — | Glob patterns to exclude |
+| `markdownIngestionExclude` | string[] | common dependency/build dirs | Glob patterns to exclude; merged with default excludes such as `node_modules/**`, `.git/**`, `dist/**`, and `build/**` |
 | `markdownIngestionDebounceMs` | number | `150` | Debounce window for file change events |
 | `markdownIngestionObsidianEnabled` | boolean | `false` | Watch Obsidian vault roots |
 | `markdownIngestionObsidianRoots` | string[] | — | Obsidian vault directories |
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -90,7 +90,7 @@ The plugin exposes `ingestionGateThreshold` for host-side gating decisions:
 | `markdownIngestionEnabled` | boolean | `false` | Watch markdown roots for changes |
 | `markdownIngestionRoots` | string[] | — | Directories to watch |
 | `markdownIngestionInclude` | string[] | — | Glob patterns to include |
-| `markdownIngestionExclude` | string[] | common dependency/build dirs | Glob patterns to exclude; merged with default excludes such as `node_modules/**`, `.git/**`, `dist/**`, and `build/**` |
+| `markdownIngestionExclude` | string[] | common dependency/build dirs | Glob patterns to exclude; merged with default excludes such as `node_modules/**`, `**/node_modules/**`, `.git/**`, `dist/**`, and `build/**` |
 | `markdownIngestionDebounceMs` | number | `150` | Debounce window for file change events |
 | `markdownIngestionObsidianEnabled` | boolean | `false` | Watch Obsidian vault roots |
 | `markdownIngestionObsidianRoots` | string[] | — | Obsidian vault directories |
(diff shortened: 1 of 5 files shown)


compoodment:
Addressed CodeRabbit's nested-directory finding in fe57ec1.

The default excludes now include both root-level and nested forms, e.g. `node_modules/**` and `**/node_modules/**`, so package/workspace subtrees like `packages/app/node_modules/` are pruned too. The integration test now covers nested `node_modules` and `build` directories and asserts no watcher is attached there.

Verification:
- `pnpm run check`
- `pnpm exec tsc -p tsconfig.tests.json && node --test .ts-build/test/integration/markdown-ingest.test.js`

– Vale

compoodment:
Self-review follow-up in latest commit: the runtime default excludes apply to both generic and Obsidian markdown adapters because they share `DirectoryMarkdownSourceAdapter`, but the PR only documented/schema-defaulted the generic `markdownIngestionExclude` field.

Fixed by adding the same default metadata/docs to `markdownIngestionObsidianExclude` too.

Verification:
- `pnpm run check`
- `pnpm exec tsc -p tsconfig.tests.json && node --test .ts-build/test/integration/markdown-ingest.test.js`

– Vale

compoodment:
## Review — Vale

### Summary
Adds default exclude patterns for common dependency/build/cache directories (`node_modules`, `.git`, `dist`, `build`, `.next`, `.venv`, `__pycache__`, etc.) to markdown ingestion. 167 additions, 6 deletions, 5 files.

### What this does well
- **Addresses a real operational gap.** Without default excludes, a workspace root like `~/projects/my-app` would recursively crawl `node_modules` and `.git`, ingesting thousands of irrelevant `.md` files from dependencies and VCS internals.
- **Merged, not replaced.** `mergeDefaultMarkdownIngestionExcludes` unions user-provided patterns with defaults using a `Set`. User patterns are additive — they can't remove defaults. This is the right approach for exclusion defaults.
- **`shouldSkipDirectory` is a good optimization.** Instead of walking into `node_modules` and then filtering each file, it skips the directory entirely during `walkDirectory`. This avoids unnecessary `stat` calls on thousands of irrelevant files.
- **Manifest defaults match runtime defaults.** The `openclaw.plugin.json` default arrays are identical to `DEFAULT_MARKDOWN_INGEST_EXCLUDES` in `markdown-ingest.ts`. Users who inspect the manifest see the actual defaults.
- **Test coverage is solid.** The integration test creates `node_modules`, `.git`, `dist`, nested `node_modules`, and nested `build`, then verifies only the legitimate doc is ingested and excluded directories are not watched.

### Concerns

1. **User cannot remove a default exclude.** If a user wants to ingest `.cache/notes.md`, they can't — `mergeDefaultMarkdownIngestionExcludes` unions user patterns with defaults, so `.cache/**` is always present. This could be addressed by supporting negation patterns (`!.cache/**`) or by allowing `null` to disable defaults, but neither is implemented here. Low severity since the current defaults cover standard build/cache directories.

2. **Duplicate patterns in manifest.** Both `markdownIngestionExclude` and `markdownIngestionObsidianExclude` have identical default arrays in `openclaw.plugin.json`. This is 46 lines of duplication. Consider extracting to a shared JSON `$ref` or a single source-of-truth constant. The runtime already uses `DEFAULT_MARKDOWN_INGEST_EXCLUDES`, so the manifest is a duplicate representation.

3. **`shouldSkipDirectory` uses `matchesGlob` which now supports `**`.** This PR merges right after #111 which upgrades `matchesGlob`. The `directoryPath` check appends `/` to the relative path: `${relative}/`. This means `node_modules/` will match `node_modules/**` and `**/node_modules/**`. Correct behavior. But `matchesGlob("node_modules", "node_modules/**")` would also match because `node_modules` doesn't contain `/` and the pattern `node_modules/**` starts with a literal match. Actually — it wouldn't match because `matchesGlob` is anchored (`^...$`). So the trailing `/` in `directoryPath` is necessary for correct matching. Good.

4. **`**/node_modules/**` and `node_modules/**` are redundant together.** `**/node_modules/**` matches `node_modules/pkg/README.md` (zero or more path segments before `node_modules`). `node_modules/**` matches the same path (one segment before `node_modules`). Having both is harmless but adds pattern evaluation cost for every directory. The `**/` prefix subsumes the top-level-only pattern.

5. **No test for `shouldSkipDirectory` directly.** The test only verifies end-to-end ingestion behavior. A unit test for `shouldSkipDirectory` with edge cases (empty relative path, nested matches) would be more targeted. Minor — the integration test is sufficient for the feature.

### Verdict
Good, practical PR. The default excludes solve a real problem. Consider allowing users to override/remove defaults in a follow-up, and deduplicating the manifest defaults. The redundant `**/` + top-level pattern pairs are harmless but could be simplified.

– Vale

xDarkicex:
Closing in favor of a minimal replacement PR. The codebase has evolved since this was authored — shouldPruneDirectory + matchesExcludedDirectory already handle the pruning logic. Only the default patterns and merge function are needed now.

PR #134, opened 2026-05-11
fix: preserve sidecar retry budget across reconnects

### LINKED ISSUE #132: test: sidecar degraded-mode retry budget expectation is broken on main
## Summary

Current `main` resets the sidecar retry counter after a successful reconnect, so the integration test that expects degraded mode after exhausting `maxRetries` no longer trips degraded mode.

That leaves the branch in a red-test state and the intended retry-budget semantics are ambiguous: either the test is stale after the retry-reset change, or the implementation now allows unlimited crash/reconnect cycles as long as each reconnect succeeds once.

## Evidence

Repo: `xDarkicex/openclaw-memory-libravdb`
Commit tested: current `main` / `v1.4.56`

Repro:

```sh
pnpm install --frozen-lockfile
pnpm run test:inspect
node --test .ts-build/test/integration/sidecar-lifecycle.test.js
```

Observed:

```text
✖ sidecar enters degraded mode after exhausting retry budget
AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:

false !== true

at .ts-build/test/integration/sidecar-lifecycle.test.js:166:12
```

Targeted run result:

```text
ℹ tests 6
ℹ pass 5
ℹ fail 1
```

The failing assertion corresponds to:

```ts
assert.equal(handle.isDegraded(), true);
```

but `handle.isDegraded()` is `false` after the second simulated crash/restart cycle.

## Likely cause

`SidecarSupervisor.start()` resets retry state on every successful start:

```ts
this.retries = 0;
```

So with `maxRetries: 1`, the test's sequence becomes:

1. crash → schedule reconnect, `retries` increments
2. reconnect succeeds → `retries` resets to `0`
3. crash again → budget appears fresh, so degraded mode is not entered

## Fix direction

Decide the intended semantics explicitly:

- If retry budget should be per continuous outage: update the integration test to assert the reset behavior.
- If retry budget should be cumulative over the supervisor lifetime/session: stop resetting `retries` on successful reconnect, or track outage retries separately from lifetime crash count.

compoodment:
## Summary

`SidecarSupervisor.start()` was resetting `this.retries = 0` on every successful reconnect, which made the retry budget appear fresh after each crash/reconnect cycle. With `maxRetries: 1`, a sidecar could crash and reconnect indefinitely without ever entering degraded mode — the integration test for this scenario was failing on main.

The fix removes the `this.retries = 0` reset from `start()`. The retry budget is now cumulative over the supervisor's lifetime: once the supervisor has scheduled `maxRetries` reconnection attempts, further crashes immediately enter degraded mode regardless of whether earlier reconnects succeeded. `this.reconnectScheduled = false` is kept as it only guards against duplicate reconnect scheduling.

## Verification

```sh
pnpm run test:inspect
node --test .ts-build/test/integration/sidecar-lifecycle.test.js
node --test .ts-build/test/unit/sidecar.test.js
```

All 6 integration tests pass (including the previously failing "sidecar enters degraded mode after exhausting retry budget"), all 11 unit tests pass.

Fixes xDarkicex/openclaw-memory-libravdb#132

– Vale

## Release Notes

* **Bug Fixes**
  * Improved connection stability and recovery behavior by ensuring proper retry state management during reconnection attempts. The system now maintains consistent tracking of connection retry cycles, preventing unexpected counter resets and enhancing overall fault tolerance capabilities during temporary network interruptions, connectivity issues, or service unavailability scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/xDarkicex/openclaw-memory-libravdb/pull/134)

### DIFF
--- a/src/sidecar.ts
+++ b/src/sidecar.ts
@@ -245,7 +245,6 @@ class SidecarSupervisor implements SidecarHandle {
   async start(): Promise<SidecarSocket> {
     const endpoint = await this.runtime.resolveEndpoint(this.cfg);
     const socket = await this.connectEndpointWithRetry(endpoint);
-    this.retries = 0;
     this.reconnectScheduled = false;
     if (this.socket instanceof SupervisorSocket) {
       this.socket.bind(socket);


xDarkicex:
Thanks for catching the flapping loophole in #98 — you were right that resetting retries immediately on every successful reconnect allows infinite crash/connect loops.

However, removing the reset entirely converts `maxRetries` from a per-outage budget into a lifetime crash counter, which permanently degrades the plugin after N total crashes over the process lifetime. A daemon that recovers from two outages over six months shouldn't trigger degraded mode on the third.

The correct fix is a **stability window** (temporal threshold): the retry counter only resets after the connection survives 15 seconds without a crash. This closes the flapping loophole (crash before the window = same outage, retries accumulate) while preserving per-outage resilience (stable connection = reset budget).

This has been implemented in #162 which supersedes this PR. Appreciate the bug report.

PR #143, opened 2026-05-12
fix: skip sidecar reconnect after shutdown

### LINKED ISSUE #136: bug: scheduled sidecar reconnect can run after shutdown
## Summary

`SidecarSupervisor.handleExit()` schedules reconnects through `runtime.scheduleRestart(...)`, but `shutdown()` only sets `shuttingDown = true` and destroys the current socket. If a reconnect was already scheduled before shutdown, the scheduled callback still calls `this.start()` later.

That means the plugin can recreate a sidecar socket after `gateway_stop` / runtime shutdown, leaving background work alive after the host expected the plugin to be stopped.

## Repro shape

1. Start the sidecar supervisor.
2. Simulate a non-zero sidecar close so `handleExit()` schedules a restart.
3. Call `shutdown()` before the restart callback runs.
4. Let the scheduled callback run.

Observed in code: the callback does not check `this.shuttingDown`, so it calls `this.start()` and creates a new socket.

## Expected

Once shutdown has begun, scheduled reconnect callbacks should be suppressed. The supervisor should not create a new sidecar socket after shutdown.

## Impact

- Plugin shutdown can leak a new daemon/socket connection.
- CLI/gateway cleanup can race with a pending reconnect and leave work running after teardown.
- Tests currently cover reconnect after crash and shutdown behavior separately, but not the pending-reconnect shutdown race.

## Suggested fix

Guard the scheduled restart callback:

```ts
this.runtime.scheduleRestart(backoffMs, () => {
  if (this.shuttingDown) {
    return;
  }
  void this.start().catch(...);
});
```

Also add a regression test that schedules a reconnect, calls `shutdown()`, runs the scheduled callback, and asserts no second socket is created.

– Vale

compoodment:
## Summary

Adds a shutdown guard to the scheduled SidecarSupervisor reconnect callback so a pending restart does not call `start()` after `shutdown()` has begun. Also adds an integration lifecycle regression test that schedules a reconnect, shuts the supervisor down, then verifies the scheduled callback does not create a new socket.

Fixes xDarkicex/openclaw-memory-libravdb#136

## Verification

- `pnpm run test:inspect` — PASS
- `pnpm exec tsc -p tsconfig.tests.json` — PASS
- `node --test --test-name-pattern="sidecar does not reconnect from a scheduled restart after shutdown" .ts-build/test/integration/sidecar-lifecycle.test.js` — PASS
- `node --test .ts-build/test/unit/*.test.js` — PASS (116 tests)
- `pnpm check` — PASS
- `npm run test:integration` — FAIL: existing upstream failure in `sidecar enters degraded mode after exhausting retry budget` (`false !== true`), reproduced unchanged on `origin/main`; appears covered by open PR #134 and this PR avoids touching that retry-budget fix.

– Vale

### DIFF
--- a/src/sidecar.ts
+++ b/src/sidecar.ts
@@ -331,6 +331,11 @@ class SidecarSupervisor implements SidecarHandle {
     this.retries += 1;
     this.reconnectScheduled = true;
     this.runtime.scheduleRestart(backoffMs, () => {
+      if (this.shuttingDown) {
+        this.reconnectScheduled = false;
+        return;
+      }
+
       void this.start().catch((error) => {
         this.reconnectScheduled = false;
         const message = error instanceof Error ? error.message : String(error);
--- a/test/integration/sidecar-lifecycle.test.ts
+++ b/test/integration/sidecar-lifecycle.test.ts
@@ -161,6 +161,24 @@ test("sidecar crash mid-session reconnects within the restart window", async () 
   assert.equal(handle.isDegraded(), false);
 });
 
+test("sidecar does not reconnect from a scheduled restart after shutdown", async () => {
+  const runtime = createRuntime({});
+  const logger = createMemoryLogger();
+  const handle = await startSidecar({ rpcTimeoutMs: 50, maxRetries: 2 }, logger, runtime.runtime);
+
+  runtime.sockets[0]?.emitClose();
+  await flushAsyncWork();
+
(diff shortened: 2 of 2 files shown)


compoodment:
Closing as a duplicate of #138. The code fix is the same, and #138 keeps the regression in the unit suite that runs under `pnpm run check`, so it is the better PR to keep.\n\n– Vale

PR #161, opened 2026-05-12
fix: preserve user turn in assembled context

### LINKED ISSUE #159: bug: assistant-only assembled context can drop the user query
## Summary

When `assemble_context_internal` or the gRPC kernel returns an assistant-only assembled replay, the plugin can forward that result as-is. Some prompt templates require at least one replay-safe user turn and fail with `No user query found in messages`.

## Scope

Context assembly/replay-shape bug. The fix must preserve a current source user turn and account for it in token budgeting; blindly appending stale user text is not acceptable.

## Fix Path

PR #161 reinserts the latest non-empty source user turn when normalized assembled output contains no user message, applies the guard to both gRPC and sidecar assemble paths, and accounts for the reinjected turn in `estimatedTokens`/budget clamping.

## Current Status

PR #161 is currently `DIRTY` and must be rebased before review.

## Verification Target

- Unit coverage for assistant-only daemon output.
- Budget-clamp coverage proving the preserved user turn survives trimming.
- Host-flow/integration coverage after rebase.

– Vale

compoodment:
## Summary

Ensures assembled context still contains a replay-safe user turn when daemon/kernel output would otherwise be assistant-only.

## Scope

- Reinserts the latest non-empty source user turn when normalized assembled output has no user message.
- Applies the guard to both gRPC kernel and sidecar RPC assemble paths.
- Accounts for the reinjected user turn in `estimatedTokens`.
- Preserves the reinjected user turn through budget clamping while reserving room for `systemPromptAddition`.

## Audit Notes

- Current merge state is `DIRTY`; this branch must be rebased before review.
- This is not the same as blindly appending stale user text. The source user turn must be current and budget-accounted.
- After rebase, rerun the focused context-engine and host-flow tests because this touches replay shape and token accounting.

## Review Follow-up

- Branch is `DIRTY`; rebase before review.
- The old #218/#221 merge-order concerns are obsolete because those PRs were closed.
- Still review the XML/system-prompt truncation path: character slicing can cut structured prompt additions mid-tag.
- Verify tool-call conversation shapes are not broken by prepending the preserved user turn.
- Prefer resilient token assertions where exact token counts are incidental.

## Verification

- `pnpm exec tsc -p tsconfig.tests.json && node --test .ts-build/test/unit/context-engine.test.js .ts-build/test/integration/host-flow.test.js`
- `pnpm run check`

Fixes #159.

– Vale

### DIFF
--- a/src/context-engine.ts
+++ b/src/context-engine.ts
@@ -164,6 +164,12 @@ function approximateMessagesTokens(messages: OpenClawCompatibleMessage[]): numbe
   return messages.reduce((sum, message) => sum + approximateMessageTokens(message), 0);
 }
 
+function hasReplaySafeUserTurn(messages: OpenClawCompatibleMessage[]): boolean {
+  return messages.some(
+    (message) => message.role === "user" && normalizeKernelContent(message.content).trim().length > 0,
+  );
+}
+
 function selectAfterTurnMessages<T>(
   messages: T[],
   prePromptMessageCount: number | undefined,
@@ -517,11 +523,9 @@ function ensureReplaySafeUserTurn(
   assembled: OpenClawCompatibleAssembleResult,
   sourceMessages: OpenClawCompatibleMessage[],
   logger?: LoggerLike,
+  tokenBudget?: number,
 ): OpenClawCompatibleAssembleResult {
-  const hasUserTurn = assembled.messages.some(
-    (message) => message.role === "user" && normalizeKernelContent(message.content).trim().length > 0,
-  );
-  if (hasUserTurn) {
+  if (hasReplaySafeUserTurn(assembled.messages)) {
     return assembled;
   }
 
@@ -533,12 +537,48 @@ function ensureReplaySafeUserTurn(
   logger?.warn?.(
     "LibraVDB assemble produced no replay-safe user turn; reinjecting the latest user message for provider compatibility.",
   );
-  const messages = [fallbackUser, ...assembled.messages];
   const baseEstimatedTokens = Math.max(
     0,
     assembled.estimatedTokens,
     approximateMessagesTokens(assembled.messages),
   );
+
+  if (typeof tokenBudget === "number" && Number.isFinite(tokenBudget) && tokenBudget > 0) {
+    const effectiveBudget = resolveEffectiveAssembleBudget(tokenBudget);
+    const fallbackCost = approximateMessageTokens(fallbackUser);
+    const fullMessages = [fallbackUser, ...assembled.messages];
+    const fullApproxTokens = fallbackCost + approximateMessagesTokens(assembled.messages);
+    if (baseEstimatedTokens + fallbackCost <= effectiveBudget && fullApproxTokens <= effectiveBudget) {
+      return {
+        ...assembled,
+        messages: fullMessages,
+        estimatedTokens: baseEstimatedTokens + fallbackCost,
+      };
+    }
+
+    if (fallbackCost >= effectiveBudget) {
+      const truncated = truncateContentToTokenBudget(fallbackUser.content, Math.max(1, effectiveBudget - 8));
+      return {
+        ...assembled,
+        messages: truncated ? [{ ...fallbackUser, content: truncated }] : [],
+        estimatedTokens: Math.min(
+          effectiveBudget,
+          truncated ? approximateMessageTokens({ ...fallbackUser, content: truncated }) : 0,
+        ),
+      };
+    }
+
+    const remainingBudget = Math.max(0, effectiveBudget - fallbackCost);
+    const trimmedMessages = trimMessagesToBudget(assembled.messages, remainingBudget);
+    const messages = [fallbackUser, ...trimmedMessages];
+    return {
+      ...assembled,
+      messages,
+      estimatedTokens: Math.min(effectiveBudget, fallbackCost + approximateMessagesTokens(trimmedMessages)),
+    };
+  }
+
+  const messages = [fallbackUser, ...assembled.messages];
   return {
     ...assembled,
     messages,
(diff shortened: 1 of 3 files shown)


compoodment:
Review finding: this PR is not merge-ready yet.

`pnpm run check` and the targeted context-engine unit suite pass, but `npm run test:integration` has a PR-specific failure in `host-flow.test.ts`:

```text
assemble passes correct configuration mapping and returns expected payload
actual includes prepended user message: { role: "user", content: "what do you remember?" }
```

The known sidecar retry-budget failure is still unrelated, but this host-flow failure is caused by this PR's behavior change. Either update the integration expectation deliberately or adjust the guard before merging.

– Vale

compoodment:
Review update: the initial host-flow expectation failure from the first review is fixed in `035648d`, but that commit still had a deeper issue — `ensureReplaySafeUserTurn()` ran before `enforceTokenBudgetInvariant()`, so a tight budget could keep an assistant-only daemon response and drop the reinjected user turn again.

Repro that fails on `035648d`:
```ts
test("reinjected user turn survives budget trimming", async () => {
  const rpc = new FakeRpc();
  rpc.assembleResponse = {
    messages: [{ role: "assistant", content: "tiny" }],
    estimatedTokens: 0,
    systemPromptAddition: "",
  };
  const engine = buildContextEngineFactory(fakeRuntime(rpc), { userId: "fixed-user" });

  const assembled = await engine.assemble({
    sessionId: "s1",
    sessionKey: "sk1",
    messages: [makeMessage("user", "this user question is long enough to be trimmed")],
    prompt: "this user question is long enough to be trimmed",
    tokenBudget: 281,
  });

  assert.equal(assembled.messages.some((message) => message.role === "user"), true);
});
```

Pushed `e907a47` so assembled output is budget-clamped first, then the latest user turn is reinserted within the remaining budget if the clamp removed it.

Validated with:
```bash
node --test .ts-build/test/unit/context-engine.test.js .ts-build/test/integration/host-flow.test.js
pnpm run check
```

`pnpm run test:integration` still fails on the unrelated sidecar retry-budget test that PR #162 fixes; I did not duplicate that sidecar fix into this branch.

compoodment:
## Review — Vale

### Summary
Ensures that `assembleContext` always includes at least one user-turn message in the output, reinjecting the latest replay-safe user message if the kernel response is assistant-only. 256 additions, 22 deletions, 3 files.

### What this does well
- **Solves a real compatibility problem.** OpenClaw providers (and most LLM APIs) require a user message in the conversation context. If LibraVDB's kernel returns only assistant messages, the provider will reject or misinterpret the request. Reinjecting the last user message is the correct fallback.
- **Budget-aware reinjection.** The `ensureReplaySafeUserTurn` function checks token budgets before reinjecting, truncates the system prompt if needed, and falls back to a minimal user-only response when the budget is extremely tight. This is thorough.
- **Warning on reinjection.** `logger?.warn?.("LibraVDB assemble produced no replay-safe user turn; reinjecting...")` makes this fallback visible in logs, which is critical for debugging.
- **Good test coverage.** Three new tests: (a) reinject when kernel returns assistant-only, (b) preserve reinjected user turn during budget clamp, (c) budget system prompt when preserving reinjected user turn. Also updates existing test expectations to account for the reinjected message.

### Concerns

1. **`findLastReplaySafeUserMessage` walks the entire `sourceMessages` array backwards.** For very long conversations (100+ messages), this is O(n) per `assemble` call. Since the function is called only when the kernel response is assistant-only (an unusual case), this is acceptable. But the `messages.some()` call in `hasReplaySafeUserTurn` also walks forward. Both are O(n) but in the typical case (kernel returns user + assistant), `hasReplaySafeUserTurn` returns true immediately and the fallback path is never taken.

2. **`truncateSystemPromptAdditionToTokenBudget` is a naive character-based truncation.** It slices `value.slice(0, maxChars)` where `maxChars = tokenBudget * APPROX_CHARS_PER_TOKEN`. This could slice mid-XML-tag or mid-word, producing malformed XML in the system prompt. The comment says "System additions are head-structured: preserve XML/preamble/instructions" but the implementation doesn't actually look for XML tag boundaries. If the system prompt contains `<context><memory>...` and truncation cuts inside `<memory>`, the result could be `<context><memo` which breaks downstream parsing.

3. **`truncateContentToTokenBudget` is called on the fallback user message** but isn't defined in this diff. It's presumably an existing utility. Verify it handles edge cases (empty content, very small budgets, multi-byte characters). Given that `maxTokensPerFile` uses `* 4 + 3` for byte estimation, the character-based approach here may not align.

4. **Reinjecting the user message changes `estimatedTokens` calculations in existing tests.** The test changes from `assert.equal(assembled.estimatedTokens, 150)` → `164`, and from `43` → `≤44`. The token count includes the reinjected user message (~8 tokens for "hello" or "what do you remember?"). This is correct but fragile — if the kernel response format changes, these exact-value assertions will break. Consider using `assert.ok(assembled.estimatedTokens <= N)` ranges instead of exact values.

5. **Interaction with PR #218 and #221.** Both of those PRs modify `context-engine.ts`'s `assemble` method. This PR wraps the result in `ensureReplaySafeUserTurn`. If all three merge, the `assemble` method becomes: kernel call → `augmentWithExactRecall` → `enforceTokenBudgetInvariant` → `ensureReplaySafeUserTurn`. PR #218 adds `takePredictiveContextInjection` after `enforceTokenBudgetInvariant`. PR #221 adds `appendPredictiveContext` after `augmentWithExactRecall`. Merge order matters — this PR needs to wrap the final result, so it should probably merge last.

6. **The reinjected user message is prepended, not appended.** The final `messages` array is `[fallbackUser, ...assembled.messages]`. This means the user message appears before the kernel's assistant messages, which is correct for conversation structure (user → assistant → user → assistant). But if the kernel returned an assistant message with tool calls, prepending a user message would change the conversation structure. Verify that this doesn't break tool-call patterns.

### Verdict
Solid, needed feature. The main risk is the naive `truncateSystemPromptAdditionToTokenBudget` which could produce broken XML. Recommend adding XML-aware truncation (find the last complete tag boundary) before merge. Watch merge order with #218/#221.

– Vale

xDarkicex:
Closing in favor of #238 — fresh implementation on current main.

PR #162, opened 2026-05-12
fix: defer sidecar retry counter reset until connection stability window expires

### LINKED ISSUE #132: test: sidecar degraded-mode retry budget expectation is broken on main

xDarkicex:
## Summary

PR #98 reset `this.retries = 0` immediately on every successful reconnect. This fixed lifetime-cumulative degradation but created a flapping vulnerability: if the daemon connects and immediately crashes, the retry counter is already back to 0, so the crash/connect loop continues forever without ever entering degraded mode.

PR #134 proposed removing the reset entirely, but that converts `maxRetries` into a lifetime crash budget — a daemon that recovers from two outages over six months degrades permanently on the third crash, even though each outage was fully recovered.

This PR implements the standard temporal-threshold pattern used by Twisted, NATS, and WebSocket reconnect libraries: retries only reset after the connection has survived a 15-second stability window.

## How it works

- `start()` schedules a deferred reset instead of immediately zeroing `retries`
- `handleExit()` clears the pending stability timer — a crash before the window means the outage is still ongoing
- `shutdown()` clears the timer as well

**Flapping**: daemon connects → crashes 50ms later → `handleExit` clears the pending reset → retries accumulate → eventually degraded. Loop closed.

**Long-term resilience**: daemon connects → stays up 15s → stability timer fires → retries reset. Per-outage budget preserved.

## Verification

```
109 unit tests: 0 fail
32 integration tests: 0 fail (including previously failing "sidecar enters degraded mode after exhausting retry budget")
```

Closes #132. Supersedes #134.

## Summary by CodeRabbit

* **New Features**
  * Supervisor now resets reconnect retries after a configurable stability window, improving recovery after intermittent connection failures.
  * New optional stabilityWindowMs configuration lets you adjust how long a connection must remain stable before retries are reset.

* **Tests**
  * Added unit tests covering endpoint resolution, backoff logic, and stability-window retry behavior to ensure reliable reconnect handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/xDarkicex/openclaw-memory-libravdb/pull/162)

### DIFF
--- a/src/sidecar.ts
+++ b/src/sidecar.ts
@@ -273,6 +273,7 @@ class SidecarSupervisor implements SidecarHandle {
       this.stabilityTimer = null;
       this.retries = 0;
     }, CONNECTION_STABILITY_WINDOW_MS);
+    this.stabilityTimer.unref?.();
   }
 
   private clearStabilityTimer(): void {
--- a/src/sidecar.ts
+++ b/src/sidecar.ts
@@ -18,6 +18,7 @@ export interface SidecarRuntime {
   resolveEndpoint(cfg: PluginConfig): string | Promise<string>;
   createSocket(endpoint: string): SidecarSocket;
   scheduleRestart(delayMs: number, restart: () => void): void;
+  stabilityWindowMs?: number;
 }
 
 class PlaceholderSocket implements SidecarSocket {
@@ -269,10 +270,16 @@ class SidecarSupervisor implements SidecarHandle {
 
   private scheduleStabilityReset(): void {
     this.clearStabilityTimer();
+    const windowMs = this.runtime.stabilityWindowMs ?? CONNECTION_STABILITY_WINDOW_MS;
+    if (windowMs <= 0) {
+      this.retries = 0;
+      return;
+    }
     this.stabilityTimer = setTimeout(() => {
       this.stabilityTimer = null;
       this.retries = 0;
-    }, CONNECTION_STABILITY_WINDOW_MS);
+      this.logger.info?.("[libravdb] sidecar connection stable; retry counter reset");
+    }, windowMs);
     this.stabilityTimer.unref?.();
   }
 
--- a/src/sidecar.ts
+++ b/src/sidecar.ts
@@ -12,6 +12,7 @@ type ErrorHandler = (error: Error) => void;
 const STARTUP_CONNECT_MAX_RETRIES = 5;
 const STARTUP_CONNECT_BASE_DELAY_MS = 100;
 const STARTUP_CONNECT_MAX_TOTAL_WAIT_MS = 2000;
+const CONNECTION_STABILITY_WINDOW_MS = 15_000;
 
 export interface SidecarRuntime {
   resolveEndpoint(cfg: PluginConfig): string | Promise<string>;
@@ -232,6 +233,7 @@ class SidecarSupervisor implements SidecarHandle {
   private degraded = false;
   private shuttingDown = false;
   private reconnectScheduled = false;
+  private stabilityTimer: ReturnType<typeof setTimeout> | null = null;
   public socket: SidecarSocket;
 
   constructor(
@@ -245,8 +247,8 @@ class SidecarSupervisor implements SidecarHandle {
   async start(): Promise<SidecarSocket> {
     const endpoint = await this.runtime.resolveEndpoint(this.cfg);
     const socket = await this.connectEndpointWithRetry(endpoint);
-    this.retries = 0;
     this.reconnectScheduled = false;
+    this.scheduleStabilityReset();
     if (this.socket instanceof SupervisorSocket) {
       this.socket.bind(socket);
     } else {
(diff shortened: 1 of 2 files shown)


compoodment (approved):
LGTM. I reviewed the current merge result against `main`, not just the PR head, because the branch is behind current main.

Local verification on the merge result:

- `pnpm install --frozen-lockfile` — PASS
- `pnpm run test:inspect` — PASS
- `pnpm exec tsc -p tsconfig.tests.json` — PASS
- `node --test .ts-build/test/integration/sidecar-lifecycle.test.js` — PASS, including `sidecar enters degraded mode after exhausting retry budget`
- `node --test .ts-build/test/unit/sidecar.test.js` — PASS
- `pnpm run check` — PASS, 116 unit tests
- `npm run test:integration` — PASS, 34 integration tests

I also ran a manual stability-window probe with `maxRetries: 1`: a reconnect that survived the 15s window got a fresh retry budget, while a second reconnect that crashed before the window degraded as expected.

Non-blocking: the PR body verification counts look stale relative to current main, and the new stability timer keeps the sidecar lifecycle test process alive until the 15s window fires. Neither blocks the fix.

– Vale

compoodment (changes requested):
Follow-up after additional review: I’m changing my verdict from clean approval to **needs small changes**.

The implementation still looks conceptually right, but two review findings are valid enough to fix before merge:

1. The 15s stability timer is currently referenced, so tests/direct sidecar handles keep Node alive until the timer expires if the handle is not explicitly shut down. I verified the minimal fix locally:

```ts
this.stabilityTimer = setTimeout(() => {
  this.stabilityTimer = null;
  this.retries = 0;
}, CONNECTION_STABILITY_WINDOW_MS);
this.stabilityTimer.unref?.();
```

With that local patch, TypeScript passes and the combined sidecar unit/integration command drops from ~17s to ~2s while still passing.

2. The PR should add coverage for the positive stability-window path: reconnect, survive past `CONNECTION_STABILITY_WINDOW_MS`, then verify a later outage receives a fresh retry budget. I manually probed this behavior and it works, but the repository tests only cover pre-window flapping/degradation. Making the window/timer injectable for tests would avoid a real 15s sleep.

No deeper correctness issue found, but these are cheap and directly tied to the new behavior.

– Vale

compoodment:
Review of current merge result against `main`:

**Fixed:** The stability timer was keeping Node alive for the full 15s window. Pushed `529919c` to `unref()` that timer. Verified locally: combined sidecar unit+integration drops from ~17s to ~2s while still passing.

**Current verification on merge result:**
- `this.stabilityTimer.unref?.();` is present
- `pnpm run test:inspect` ✅
- `pnpm run check` ✅ — 117/117
- `npm run test:integration` ✅ — 34/34

**Remaining blocker — coverage, not runtime behavior:**
The PR still only changes `src/sidecar.ts`. It needs deterministic regression coverage for the temporal retry reset it introduces, at minimum:
1. Crash before the 15s stability window preserves/increments retry count and can degrade
2. Surviving past the window resets the retry count
3. Shutdown clears the pending stability timer

Without that, the exact flapping bug this PR fixes can regress silently.

— Vale

PR #174, opened 2026-05-13
fix: gate memory readFile paths to search results

IWhatsskill:
## Summary

Tighten the memory runtime bridge so `readFile()` only opens memory paths that were returned by the same search manager's structured `search()` results.

The normal search path already scopes collections before returning results. The follow-up read path was looser: it decoded the caller-provided `collection::id` path and called `list_collection` for that decoded collection without checking whether that path came from a prior structured search result.

## Root cause

`readFile({ relPath })` treated the encoded path as authority. That meant a caller could provide a crafted memory path shape and make the bridge ask the daemon for that decoded collection. The daemon still decides what exists, but the bridge was not preserving the search result boundary it had just established.

## What changed

- Track paths returned by structured `search()` results per search manager.
- Reject `readFile()` calls when the path was not returned by that manager.
- Keep legacy string search behavior unchanged.
- Avoid granting hidden read access from legacy search results, since those do not return structured paths.
- Add regression coverage for crafted paths and legacy hidden paths.

## Why this is safe

- Existing structured search results still round-trip through `readFile()`.
- Crafted paths are rejected before any `list_collection` call is made.
- The change is local to the memory runtime bridge and tests.
- No daemon protocol, dependency, network, or configuration behavior changes.

## Tests

- `node_modules/.bin/tsc.CMD -p tsconfig.tests.json`
- `node --test .ts-build/test/unit/memory-runtime.test.js`

Also checked:

- `node --test .ts-build/test/unit/*.test.js`
  - 131/132 passed
  - Existing unrelated Windows path expectation failed in `test/unit/sidecar.test.ts`: `defaultEndpoint prefers the Homebrew socket when the user-local socket is absent`

## Not run

- `npm run test:integration`, because the integration tests require a daemon/prepared daemon binary.

### DIFF
--- a/src/memory-runtime.ts
+++ b/src/memory-runtime.ts
@@ -67,6 +67,7 @@ function createMemorySearchManager(
 ) {
   let cachedStatus = initialStatus;
   let cachedIdentityUserId: string | null = null;
+  const returnedSearchPaths = new Set<string>();
 
   function getResolvedUserId(sessionKey: string | undefined): string {
     if (cachedIdentityUserId !== null) return cachedIdentityUserId;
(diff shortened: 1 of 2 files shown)


compoodment (approved):
Looks good. The new per-manager authorization set closes the readFile gap at the right seam: structured searches authorize only the exact paths returned by that manager, crafted collection::id paths are rejected before list_collection, and legacy string search remains unable to smuggle hidden read access.

I checked the surrounding runtime bridge and did not find a compatibility blocker. Keeping authorization manager-local is the right scope here; it preserves existing structured readFile behavior while avoiding a global path cache.

Verification run locally on the PR branch:
- COREPACK_ENABLE_AUTO_PIN=0 pnpm run check — pass
- COREPACK_ENABLE_AUTO_PIN=0 pnpm run test:integration — pass, 40/40

No blockers from me.

– Vale

PR #175, opened 2026-05-13
fix: build unix endpoint paths with posix separators

IWhatsskill:
## Summary

Make `defaultEndpoint()` build Unix socket paths with POSIX separators after the requested platform has already been classified as non-Windows.

This keeps the Windows TCP endpoint unchanged, while making simulated macOS/Linux endpoint discovery host-independent in the unit suite.

## Root cause

`defaultEndpoint(platform, homeDir, pathExists)` accepts an explicit platform argument, so tests can simulate Darwin behavior on another host OS. The implementation still used the host `path.join()` for Unix socket candidates. When the test suite runs on Windows and asks for `platform === "darwin"`, that produced backslash paths such as `\Users\demo\.libravdbd\run\libravdb.sock`, so the Homebrew socket candidate check could not match the expected POSIX path.

## What changed

- Keep the existing Windows branch returning `tcp:127.0.0.1:37421`.
- Use POSIX path joining for non-Windows Unix socket candidates and fallback paths.
- Add a regression test for Darwin fallback path formatting on a Windows-hosted test run.

## Why this is safe

- Runtime Windows behavior is unchanged.
- POSIX hosts already expect `/` separators for Unix sockets, so the produced endpoint shape stays correct.
- The change only affects local endpoint string construction.
- No daemon lifecycle, install, network, or config contract behavior changes.

## Tests

- `.\node_modules\.bin\tsc.CMD -p tsconfig.tests.json`
- `node --test .ts-build\test\unit\sidecar.test.js`
- `node --test .ts-build\test\unit\*.test.js`

## Not run

- Daemon/integration tests, because this is a path-string unit fix and does not start or talk to a daemon.

## Notes

- Full unit suite passes on Windows: 132/132.

### DIFF
--- a/src/sidecar.ts
+++ b/src/sidecar.ts
@@ -439,10 +439,11 @@ export function defaultEndpoint(
     return "tcp:127.0.0.1:37421";
   }
 
+  const joinSocketPath = path.posix.join;
   const sockName = "libravdb.sock";
   const candidateDirs = [
     // User-local (npm plugin convention)
-    homeDir?.trim() ? path.join(homeDir, ".libravdbd", "run") : null,
+    homeDir?.trim() ? joinSocketPath(homeDir, ".libravdbd", "run") : null,
     // Homebrew (Apple Silicon) — matches the Homebrew formula LaunchAgent
     "/opt/homebrew/var/libravdbd/run",
     // Homebrew (Intel Mac) / manual Linux installs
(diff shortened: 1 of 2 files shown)


compoodment (approved):
Looks good. Using path.posix.join only after the platform has already been classified as non-Windows fixes the host-dependent Windows test failure without changing the Windows TCP endpoint path. The Homebrew/user-local probe order stays intact, and the fallback endpoint now matches the Unix socket convention regardless of the test host OS.

Verification run locally on the PR branch:
- COREPACK_ENABLE_AUTO_PIN=0 pnpm run check — pass, 132/132 unit tests
- COREPACK_ENABLE_AUTO_PIN=0 pnpm run test:integration — pass, 40/40

No blockers from me.

– Vale

PR #176, opened 2026-05-13
fix: validate user memory collections

IWhatsskill:
## Summary

Validate user-scoped memory collection names through a shared helper before sending probe/search calls to the daemon.

PR #129 added validation for `resolveDurableNamespace()`. A few follow-up paths still built `user:${userId}` directly, which meant malformed config/framework/identity user IDs could still become daemon collection names outside that helper.

## Root cause

The validation existed in one namespace-resolution path, but user-scoped collection construction was duplicated in other paths:

- exact recall probes in the context engine
- memory runtime user/global searches
- `memory status --deep` health probes

Those paths were simple and easy to miss because the happy-path value is the same string shape: `user:<id>`.

## What changed

- Add `resolveUserCollection()` in `memory-scopes`.
- Validate both the raw user namespace and the final `user:<id>` collection name.
- Use the helper for exact recall probes before calling `search_text_collections`.
- Use the helper for memory runtime user/global searches.
- Use the helper for `memory status --deep`.
- When `status --deep` sees an invalid user collection, report that probe as failed and skip only that user collection instead of sending it to the daemon.
- Add focused unit coverage for invalid user IDs in the shared helper, context engine, and CLI deep status path.

## Why this is safe

- Valid existing user IDs continue to produce the same `user:<id>` collection names.
- Invalid user-scoped collection names fail before daemon calls.
- `status --deep` still checks authored collections and `global` even when the configured user ID is invalid.
- The change reuses the same namespace rules introduced by #129.
- No daemon protocol, dependency, network, install, or persisted data format changes.

## Tests

- `.\node_modules\.bin\tsc.CMD -p tsconfig.tests.json`
- `node --test .ts-build\test\unit\memory-scopes.test.js`
- `node --test .ts-build\test\unit\context-engine.test.js`
- `node --test .ts-build\test\unit\cli.test.js`
- `node --test .ts-build\test\unit\memory-runtime.test.js`

## Not run

- Daemon/integration tests, because this is plugin-side validation and unit coverage exercises the affected RPC call boundaries.

## Notes

- Full unit suite on the current `main` baseline still has the known unrelated Windows `defaultEndpoint` sidecar failure fixed separately by #175.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened validation for user collection names; invalid namespaces are now reported and excluded from searches, with fallback to global collections.

* **Refactor**
  * Collection resolution and cross-session recall now use normalized/validated user collection names for more consistent lookup behavior.

* **Tests**
  * Added unit tests covering invalid namespace handling, error reporting, and the updated collection-resolution/search behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/xDarkicex/openclaw-memory-libravdb/pull/176)

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -2,7 +2,7 @@ import { createInterface } from "node:readline/promises";
 import { stdin, stdout } from "node:process";
 import type { OpenClawPluginApi } from "openclaw/plugin-sdk/plugin-entry";
 import { MEMORY_CLI_DESCRIPTOR, isMemorySlotSelected } from "./cli-descriptors.js";
-import { resolveDurableNamespace } from "./memory-scopes.js";
+import { resolveDurableNamespace, resolveUserCollection } from "./memory-scopes.js";
 import { resolveIdentity } from "./identity.js";
 import { formatError } from "./format-error.js";
 import { promoteDreamDiaryFile } from "./dream-promotion.js";
@@ -324,11 +324,21 @@ async function runDeepStatusProbe(
     identityPath: cfg.identityPath,
     noAutoPersist: true,
   });
-  const durableCollections = [`user:${userId}`, "global"] as const;
-
-  const allCollections = [...AUTHORED_STATUS_COLLECTIONS, ...durableCollections] as const;
-
   const probes: DeepStatusProbe[] = [];
+  let userCollection: string | null = null;
+  try {
+    userCollection = resolveUserCollection(userId);
+  } catch (error) {
+    probes.push({
+      ok: false,
+      collection: "user:<invalid>",
+      error: formatError(error),
+    });
+  }
+
+  const durableCollections = userCollection ? [userCollection, "global"] : ["global"];
+  const allCollections = [...AUTHORED_STATUS_COLLECTIONS, ...durableCollections];
+
   for (const collection of allCollections) {
     try {
       const result = await rpc.call<{ results?: unknown[] }>("search_text", {
--- a/src/context-engine.ts
+++ b/src/context-engine.ts
@@ -13,6 +13,7 @@ import {
   CompactSessionResponse,
 } from "@xdarkicex/libravdb-contracts";
 import { resolveIdentity, type ResolvedIdentity } from "./identity.js";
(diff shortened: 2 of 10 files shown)


compoodment (changes requested):
Found one blocker. The new helper closes the `user:` paths, but the dream-search branch still builds a user-derived daemon collection without validating the final prefixed name.

`createMemorySearchManager.search()` validates the raw durable namespace, then for dream queries calls `search_text` with `resolveDreamCollection(userId)`. `resolveDreamCollection()` currently just returns `dream:${userId.trim()}`. That means a raw user id that passes namespace validation can still become an invalid daemon collection after the `dream:` prefix is added. Concrete edge case: 123 `a` characters is a valid raw namespace and a valid `user:` collection length, but `dream:` + that id is 129 chars, exceeding the collection regex max, and this path still sends it to the daemon.

That leaves the same prefixed user-derived collection-name gap this PR is trying to close, just on dream-routed memory searches. Please validate the final dream collection name as well, preferably via a dedicated `resolveDreamCollection()` validation helper using the same namespace rules, and add focused memory-runtime/dream-routing coverage that proves the invalid dream collection is rejected before `search_text`.

Checks I ran:
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm run check` — pass, 134/134
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm run test:integration` — pass, 40/40
- Merged with #174 and #175 locally — no conflicts
- Combined `COREPACK_ENABLE_AUTO_PIN=0 pnpm run check` after that stack — pass, 137/137

– Vale

IWhatsskill:
@compoodment re-review pls

compoodment (approved):
Re-reviewed after the follow-up. The previous blocker is fixed: `resolveDreamCollection()` now validates both the raw namespace and the final `dream:` collection name, and the memory-runtime test proves the oversized prefixed dream collection is rejected before `search_text` reaches the daemon.

I did not find any remaining blockers in this PR.

Checks I ran:
- `pnpm run check` — pass, 135/135
- `npm run test:integration` — pass, 40/40

Approved.

– Vale

Issue #177, opened 2026-05-13
bug: fresh markdown ingestion can peg all CPU cores


compoodment:
## Summary

Fresh markdown ingestion can drive `libravdbd` to sustained all-core CPU even with a small authored-markdown root.

## Evidence

Observed with plugin v1.4.69 and daemon v1.4.62/v1.4.64 against a clean DB and 35 included markdown files after generated paths were pruned.

```text
v1.4.64: daemon reached ~=1550% CPU within the first 11s
v1.4.62: daemon reached ~=1470% CPU within the first 11s
v1.4.62 extended sample: daemon remained ~=1480-1527%
```

## Scope

This appears daemon-side. The plugin queues markdown files sequentially, so JS-side request fan-out is not the obvious cause.

## Expected Behavior

Startup markdown ingestion should have bounded CPU/backpressure controls by default.

## Fix Path

- Daemon-side embedding/tokenizer concurrency controls.
- Daemon-side request backpressure.
- Optional plugin pacing as mitigation, not the primary fix.

## Verification Target

- Bounded repro with clean DB, known file count, and CPU samples.
- Confirm whether newer daemon releases still show saturation before changing plugin behavior.

– Vale

xDarkicex:
Verified stale: the markdown ingestion system now has sophisticated backpressure and streaming controls.\n\nThe `IngestQueue` class (src/ingest-queue.ts) implements:\n- Token-based chunking with sentence-boundary awareness\n- Exponential backoff with full jitter retry\n- Daemon feedback-driven adaptive token burst limiting (`tokenBurstLimit`)\n- `acceptMore` backpressure signal respected from daemon\n- Chunk-level feedback reporting with `queueDepth`/`queueCapacity`\n- DELETE queue with retry semantics\n\nThe daemon controls `tokenBurstLimit` and `acceptMore` signals that the plugin respects. This is a complete backpressure system. Issue was daemon-side focused and has been addressed by the daemon's feedback protocol.

PR #178, opened 2026-05-13
fix: pace markdown startup ingestion

compoodment:
## Summary

- pace startup markdown ingestion by limiting changed/new files processed per scan
- cap changed/new markdown bytes read and ingested per scan so large notes cannot bypass the file-count throttle
- continue deferred generic and Obsidian scans after a configurable delay
- add config/schema/docs for per-scan file caps, byte caps, and batch delays (`0` disables a cap)
- log deferred counts and cover continuation/read-deferral behavior with integration tests

## Context

Fresh markdown startup scans can saturate daemon CPU while embedding/tokenizing documents. This PR bounds both daemon ingest request rate and plugin-side read/hash/decode work during startup scans. Daemon-side CPU/thread/backpressure controls are tracked separately in #177.

## Tests

- `npm run test:integration -- --test-name-pattern="markdown ingestion"`
- `pnpm run check`
- `npm run test:integration`
- `npm run build`

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -92,11 +92,15 @@ The plugin exposes `ingestionGateThreshold` for host-side gating decisions:
 | `markdownIngestionInclude` | string[] | — | Glob patterns to include |
 | `markdownIngestionExclude` | string[] | — | Glob patterns to exclude |
 | `markdownIngestionDebounceMs` | number | `150` | Debounce window for file change events |
+| `markdownIngestionMaxFilesPerScan` | number | `4` | Maximum changed/new files to ingest per scan; `0` disables the cap |
+| `markdownIngestionBatchDelayMs` | number | `1000` | Delay before continuing a scan with deferred files |
 | `markdownIngestionObsidianEnabled` | boolean | `false` | Watch Obsidian vault roots |
 | `markdownIngestionObsidianRoots` | string[] | — | Obsidian vault directories |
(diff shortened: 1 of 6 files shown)


xDarkicex:
Thanks for the work on this — the problem is real and well-scoped. Fresh scans saturating daemon CPU during startup is a genuine issue that needs solving.

However, the current approach (fixed FIFO caps on files and bytes) is a blunt instrument that doesn't address the underlying mechanics:

- No mtime/ctime priority — fresh files and stale files are processed equally, so deferred files are often the least relevant ones
- Size-based budgeting doesn't map to token cost — a 10-byte note and a 500KB vault dump count equally against the byte cap
- No token-aware budget — embedding cost is proportional to tokens, not bytes
- No daemon feedback loop — the plugin paces blind; if the daemon is under load, there's no signal to slow down
- Oversized file loophole — a single 100MB file can still saturate the daemon in one scan

A proper solution would include priority-based ordering (newest files first), token estimation before committing budget, size-tiered batching, and daemon latency-based backpressure. We've documented this in `startup_markdown_ingestion.md` as a reference for a v2 implementation.

The PR establishes a blunt cap as the documented "solution" and adds it to the config schema, which would make a proper refactor harder to justify later. For that reason, I'm closing this and recommending the more structured approach instead.

Thanks again for identifying and investigating the issue.

PR #183, opened 2026-05-15
Codex/fix sidecar retry loop

JuanHuaXu:
Reconnect retry loop only tries one scheduled restart after a crash. src/sidecar.ts:330-338 increments retries and schedules start(), but if that reconnect fails, the catch only logs and clears reconnectScheduled. Nothing calls handleExit() again, so maxRetries is not actually consumed beyond the startup retry loop inside that one attempt. A daemon that comes back slightly later can leave the plugin stuck until RPC calls time out.

## Release Notes

* **Bug Fixes**
  * Enhanced sidecar reconnection failure handling to correctly manage retry mechanisms and degraded mode transitions, improving reliability during intermittent connection issues.

[Review Change Stack](https://app.coderabbit.ai/change-stack/xDarkicex/openclaw-memory-libravdb/pull/183)


compoodment (approved):
Local gate:
`pnpm run check` passes on `3e5fca1`.
Focused sidecar lifecycle test also passes 7/7.

The fix is the right shape: failed scheduled reconnects now re-enter `handleExit(1)`, so the retry budget advances instead of logging once and getting stuck. Covered edge cases: retry scheduling, degraded transition after budget exhaustion, successful reconnect after transient failures, and preserving the supervisor socket path for existing RPC callers. No blocker found.

– Vale

PR #184, opened 2026-05-15
Codex/fix grpc auth init fail closed

JuanHuaXu:
gRPC auth initialization failure is swallowed, then authenticated calls fail later. src/context-engine.ts:875-885 catches initializeSession() errors and proceeds. If LIBRAVDB_AUTH_SECRET is set and nonce negotiation fails, the next call throws call initializeSession before authenticated RPCs; bootstrap/assemble then falls back or fails in a less useful place. Auth failure should fail closed with a direct diagnostic.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for gRPC authentication initialization failures during system startup, providing clearer error messages when authentication configuration issues occur.

* **Tests**
  * Added unit tests to validate gRPC authentication initialization failure handling and error messaging.

[Review Change Stack](https://app.coderabbit.ai/change-stack/xDarkicex/openclaw-memory-libravdb/pull/184)


JuanHuaXu:
@coderabbitai review

compoodment (approved):
Local gate:
`pnpm run check` passes on `ece0d80`.
Focused context-engine unit test passes 38/38.

The behavior is correct for the threat model: when auth is explicitly configured, swallowing `InitializeSession` failures is unsafe because later authenticated RPCs can fail in less obvious places. Failing closed with a sanitized error is better. The no-auth path still preserves the optional-init fallback, so this should not break unauthenticated/local deployments.

– Vale

PR #188, opened 2026-05-15
intelligent markdown startup ingestion (priority/token/tier/stream)

xDarkicex:
## Summary

Implements intelligent markdown startup ingestion pacing with priority queuing, token-aware budgeting, size-tiered batching, and bounded streaming reads.

## What's done

**Priority queue** — files are sorted before processing. Default is mtime (newest first), also supports ctime, size, and fifo. Deferred files are always the oldest/stalest.

**Token budget** — per-scan budget in tokens (default 8192). Estimates tokens as `size / 4`. When budget is exhausted, remaining files are deferred.

**Size-tiered batching** — per-cycle limits by tier:
- Small (<10KB): max 16 per cycle
- Medium (10KB–100KB): max 4 per cycle
- Large (>100KB): max 1 per cycle

**Max tokens per file** — hard cap (default 128K tokens). Files over this are deferred without reading.

**Streaming reads** — files read in 64KB chunks with incremental FNV-1a64 hash. Bounded memory regardless of file size. Max bytes derived from maxTokensPerFile.

**Config fields** (in `types.ts`, `openclaw.plugin.json`, docs):
- `markdownIngestionPriorityMode` — sort mode: mtime/ctime/size/fifo
- `markdownIngestionTokenBudgetPerScan` — tokens per scan (0=unlimited)
- `markdownIngestionSmallFileThreshold` — small file byte threshold
- `markdownIngestionLargeFileThreshold` — large file byte threshold
- `markdownIngestionMaxTokensPerFile` — max tokens per single file

## What's not done (Phase 5 — separate effort)

**Daemon backpressure** — requires daemon RPC contract changes to return ingest duration metadata. Out of scope here.

**Deprecated config cleanup** — old fields (MaxBytesPerScan, MaxFilesPerScan, BatchDelayMs) from PR #178 are not explicitly handled.

## Tests

- mtime priority ordering
- token budget deferral
- oversized file deferral
- size-tier enforcement
- streaming read boundedness

## Summary by CodeRabbit

* **New Features**
  * Configurable markdown ingestion: priority mode (mtime/ctime/size/fifo) and per-file token limit.
  * Streaming-safe ingestion with adaptive chunking that responds to daemon feedback and resumes scans under backpressure.
  * Per-file deferral/skipping for oversized files and resume cursor behavior for interrupted scans.

* **Tests**
  * Extended integration/unit tests covering startup ordering, backpressure/resume, streaming reads, and feedback-driven behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/xDarkicex/openclaw-memory-libravdb/pull/188?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/openclaw.plugin.json
+++ b/openclaw.plugin.json
@@ -263,6 +263,20 @@
         "type": "number",
         "default": 150
       },
+      "markdownIngestionPriorityMode": {
+        "type": "string",
+        "enum": [
+          "mtime",
+          "ctime",
+          "size",
+          "fifo"
+        ],
+        "default": "mtime"
+      },
+      "markdownIngestionMaxTokensPerFile": {
+        "type": "number",
+        "default": 128000
+      },
       "markdownIngestionSnapshotPath": {
         "type": "string"
       },
--- a/src/ingest-queue.ts
+++ b/src/ingest-queue.ts
@@ -131,6 +131,14 @@ export class IngestQueue {
         continue;
       }
 
+      if (lastFeedback && lastFeedback.nodesAccepted === 0) {
+        this.logger.warn?.(
+          `[ingest-queue] Chunk permanently rejected for ${sourceDoc} ` +
+          `at offset=${offset} length=${chunkText.length} ` +
+          `tokenBurstLimit=${lastFeedback.tokenBurstLimit ?? "unset"}`,
+        );
+      }
+
       if (this.options.onChunkFeedback && lastFeedback) {
         this.options.onChunkFeedback(lastFeedback);
       }
--- a/src/ingest-queue.ts
+++ b/src/ingest-queue.ts
@@ -8,6 +8,8 @@ export interface IngestQueueOptions {
   retryBaseDelayMs: number;
   /** Max retries per chunk. */
   maxRetries: number;
+  /** Called after each chunk is accepted so scan-level state stays current. */
+  onChunkFeedback?: (feedback: IngestFeedback) => void;
 }
 
 const DEFAULT_OPTIONS: IngestQueueOptions = {
@@ -15,13 +17,6 @@ const DEFAULT_OPTIONS: IngestQueueOptions = {
   retryBaseDelayMs: 500,
   maxRetries: 4,
 };
-
-interface QueuedIngest {
-  sourceDoc: string;
-  params: IngestMarkdownDocumentParams;
-  resolve: () => void;
-  reject: (err: Error) => void;
-}
 
 interface IngestMarkdownDocumentParams {
   sourceDoc: string;
@@ -35,10 +30,37 @@ interface IngestMarkdownDocumentParams {
     fileHash: string;
     sourceSize: number;
     sourceMtimeMs: number;
+    sourceCtimeMs: number;
     ingestVersion: number;
     hashBackend: string;
   };
   mode?: IngestMode;
+}
+
+interface IngestFeedback {
+  queueDepth: number;
+  queueCapacity: number;
+  acceptMore: boolean;
+  retryAfterMs: number;
+  processingTimeUs: number;
+  nodesAccepted: number;
+  nodesRejected: number;
+  tokensIngested: number;
+  tokenBurstLimit: number;
+  walDepth?: number;
+  walCapacity?: number;
+}
+
+interface IngestMarkdownDocumentResponse {
+  ok: boolean;
+  feedback?: IngestFeedback;
+}
+
+interface QueuedIngest {
+  sourceDoc: string;
+  params: IngestMarkdownDocumentParams;
+  resolve: () => void;
+  reject: (err: Error) => void;
 }
 
 export class IngestQueue {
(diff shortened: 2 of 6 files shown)


compoodment:
Review result: request changes.

I ran this in a detached review worktree at `13d1aa5`.

Gates:
- `pnpm run check`: passed
- `pnpm run plugin:ci`: passed with the existing inspector-gap report
- `pnpm exec tsc -p tsconfig.tests.json && node --test .ts-build/test/integration/markdown-ingest.test.js`: failed

The touched markdown ingestion integration suite currently fails:

```
✖ markdown ingestion startup defers oversized files and stops when scan token budget is depleted
AssertionError: false !== true
```

Blockers:

1. `src/markdown-ingest.ts:472` can create a tight rescan loop. If a candidate is larger than `tokenBudgetPerScan` but still under `maxTokensPerFile`, every scan hits this branch, schedules another scan, and makes no progress. I reproduced this with one markdown file: 15 scans in 25ms, all `deferred=1`, `ingested=0`.

2. `src/markdown-ingest.ts:477` tier-deferred files are counted as deferred but no follow-up scan is scheduled. Once a tier cap is hit, later eligible files can remain unindexed until some unrelated watcher event happens.

3. The failing test exposes an ordering bug: `fifo` is based on `readdir` traversal order, not creation/write order. In my run `medium.md` was evaluated before `small.md`, so the budget branch broke the scan before ingesting the small file.

4. The new config fields are missing from `openclaw.plugin.json`:
   - `markdownIngestionPriorityMode`
   - `markdownIngestionTokenBudgetPerScan`
   - `markdownIngestionSmallFileThreshold`
   - `markdownIngestionLargeFileThreshold`
   - `markdownIngestionMaxTokensPerFile`

   Since the manifest has `additionalProperties: false`, users cannot configure the new knobs through normal plugin config.

CodeRabbit's comments are valid too:
- the new stat/stream helpers bypass the injected `FsApi` seam via direct `fsp.stat` / `fsp.open`
- the mtime test should use explicit timestamps rather than a 10ms delay

xDarkicex:
token budget will not function properly until I update vector service to relay budget and backoff data.

compoodment:
## Review — Vale

### Summary
Overhauls markdown ingestion to use priority-based file ordering, per-scan token budgets, file size tiers, and streaming file reads. 310 additions, 22 deletions, 3 files.

### What this does well
- **Priority-based scan ordering.** Files are collected into `FileCandidate[]`, sorted by `priorityMode` (`mtime`, `ctime`, `size`, `fifo`), then ingested in priority order. Newest-first (`mtime`) is the right default — it processes recently changed docs first.
- **Token budget per scan.** `tokenBudgetPerScan` (default 8192) caps how many estimated tokens are ingested in a single startup scan. This prevents a huge workspace from blocking startup while ingesting thousands of files.
- **Tiered size filtering.** `allowTier` enforces that small files (< 10KB) always get through, medium files (10-100KB) are limited to 4 per scan, and large files (> 100KB) each take a tier slot. This ensures small docs aren't starved by a few large files.
- **Streaming reads via `safeReadFileStreamed`.** Instead of `readFile`-ing entire files into memory, this uses chunked reads with a `STREAM_CHUNK_BYTES` (64KB) buffer and a `maxBytes` cap. Files exceeding `maxTokensPerFile * 4 + 3` bytes are marked `too_large` and skipped. This is a real improvement for large workspaces.
- **FNV-1a streaming hash.** Instead of `hashBytes(bytes)`, the stream reader updates an FNV-1a 64-bit hash incrementally. This avoids buffering the entire file for hashing.
- **Good test coverage.** Tests verify mtime ordering, token budget depletion, and streaming reads (via a `readFile` mock that throws, proving the stream path is used).

### Concerns

1. **`estimateTokens` is not defined in the diff.** The function is used in `syncCandidates` but the diff doesn't show its definition. Assuming it's `size / 4` or similar, but verify it accounts for multi-byte characters. CJK text can be 3-4 bytes per character but 1-2 tokens per character, so `size / 4` may underestimate.

2. **FNV-1a implementation uses `BigInt.asUintN(64, ...)`.** BigInt arithmetic per byte is slow. For a 100KB file, that's 100K iterations of `BigInt.asUintN(64, hash * prime)`. On a V8 engine this is 5-10x slower than the `hashBytes` it replaces (which operates on `Uint8Array` with number arithmetic). For small files this doesn't matter, but for files approaching the `largeFileThreshold` (100KB), this could be noticeable. Consider using a `DataView` + number-based FNV-1a for the lower 32 bits and a separate upper 32-bit carry, or just use the existing `hashBytes` for files that fit in memory and fall back to streaming only for large files.

3. **`STREAM_CHUNK_BYTES = 64 * 1024` is a magic number.** This PR is supposed to be extracting magic numbers, but it introduces a new one. Should be `DEFAULT_STREAM_CHUNK_BYTES` or `STREAM_READ_CHUNK_BYTES` to match the project's naming convention from #214.

4. **`allowTier` has a hard-coded limit of 4 medium-tier files.** The `4` is a magic number. If a workspace has 5 medium-sized docs (10-100KB each), the 5th will be deferred. This is probably fine for default behavior, but the limit should be configurable or at least a named constant.

5. **`stats.filesDeferred++` is used for both "over token budget" and "over tier limit".** These are different deferral reasons. A monitoring dashboard or log analysis tool can't distinguish "deferred because budget exhausted" from "deferred because tier limit hit". Consider splitting into `filesDeferredBudget` and `filesDeferredTier`.

6. **`syncMarkdownFile` now takes an `initialStat` optional parameter.** This is a good optimization — the stat was already computed during candidate collection. But `initialStat` only includes `size` and `mtimeMs`, not the full `fs.Stat` object. If `syncMarkdownFile` needs other stat fields (e.g., `ctimeMs` for cache validation), it falls back to `safeStat(filePath)`. Verify that the fallback path works correctly.

7. **The `priorityMode` config key isn't in the PR's manifest changes.** I see `markdownIngestionPriorityMode` added to `types.ts`, but the diff doesn't show the `openclaw.plugin.json` configSchema update. If the key isn't in the schema, it won't be validated or documented. Same for `markdownIngestionTokenBudgetPerScan`, `markdownIngestionSmallFileThreshold`, `markdownIngestionLargeFileThreshold`, and `markdownIngestionMaxTokensPerFile`. These should all be in `configSchema.properties`.

8. **Interaction with PR #131 (default excludes).** #131 adds `shouldSkipDirectory` which skips excluded directories during `walkDirectory`. This PR changes `walkDirectory` to also collect `FileCandidate[]`. If #131 merges first, `shouldSkipDirectory` will prevent `node_modules` directories from appearing in candidates. If this PR merges first, candidates could include files from excluded directories. Merge order: #131 should go first.

### Verdict
Substantial, well-structured improvement to ingestion performance. Main risks: (a) BigInt FNV-1a performance, (b) missing configSchema entries for new options, (c) merge order dependency with #131. The magic numbers (`STREAM_CHUNK_BYTES`, tier limit `4`) should be named constants per #214's recommendation.

– Vale

PR #190, opened 2026-05-15
Malformed RPC response frames create ambiguous mutating-call outcomes.

JuanHuaXu:
Malformed RPC response frames create ambiguous mutating-call outcomes. src/rpc.ts:189-219 catches protobuf decode errors and waits for the pending call to time out. For writes such as ingest, the daemon may have committed the request before emitting a malformed response, so the client reports a timeout with unknown commit status. Retrying can duplicate work unless the daemon layer is perfectly idempotent. The client should fail fast with a protocol error by rejecting pending calls or closing the socket.

## Summary by CodeRabbit

* **Bug Fixes**
  * RPC client now treats malformed or unsupported response frames as protocol violations, closing the connection with a clear error instead of ignoring them.
  * Connections are terminated on invalid response shapes (e.g., bad message IDs), improving detection and debugging of communication errors.

* **Tests**
  * Added/updated tests to verify protocol-violation handling and socket error propagation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/xDarkicex/openclaw-memory-libravdb/pull/190)

### DIFF
--- a/src/rpc.ts
+++ b/src/rpc.ts
@@ -192,8 +192,8 @@ export class RpcClient {
     try {
       const msg = RpcResponse.fromBinary(payload as Uint8Array);
 
-      if (typeof msg.id !== "bigint") {
-        this.socket.destroy(new Error("Protocol violation: expected bigint message id"));
+      if (typeof msg.id !== "bigint" || msg.id <= 0n) {
+        this.socket.destroy(new Error("Protocol violation: expected positive bigint message id"));
         return false;
       }
 
--- a/src/rpc.ts
+++ b/src/rpc.ts
@@ -174,7 +174,9 @@ export class RpcClient {
       const payload = this.rxBuf.subarray(4, frameSize);
       this.rxBuf = this.rxBuf.subarray(frameSize);
 
-      this.dispatchMessage(payload);
+      if (!this.dispatchMessage(payload)) {
+        return;
+      }
     }
 
     // Compaction guard: release large backing allocations if the remainder is tiny
@@ -186,18 +188,18 @@ export class RpcClient {
     }
   }
 
-  private dispatchMessage(payload: Buffer): void {
+  private dispatchMessage(payload: Buffer): boolean {
     try {
       const msg = RpcResponse.fromBinary(payload as Uint8Array);
 
       if (typeof msg.id !== "bigint") {
         this.socket.destroy(new Error("Protocol violation: expected bigint message id"));
-        return;
+        return false;
       }
 
       const pending = this.pending.get(msg.id);
       if (!pending) {
-        return;
+        return true;
       }
 
       clearTimeout(pending.timer);
@@ -206,7 +208,7 @@ export class RpcClient {
       if (msg.error) {
         const message = msg.error.message?.trim() || `RPC error ${msg.error.code}`;
         pending.reject(new Error(msg.error.code ? `${message} (${msg.error.code})` : message));
-        return;
+        return true;
       }
 
       try {
@@ -214,8 +216,11 @@ export class RpcClient {
       } catch (error) {
         pending.reject(error instanceof Error ? error : new Error(String(error)));
       }
-    } catch {
-      // Ignore malformed frames
+      return true;
+    } catch (error) {
+      const message = error instanceof Error ? error.message : String(error);
+      this.socket.destroy(new Error(`Protocol violation: malformed RPC response frame: ${message}`));
+      return false;
     }
   }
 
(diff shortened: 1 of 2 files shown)


JuanHuaXu:
@coderabbitai review

compoodment (commented):
One remaining malformed-frame path still needs handling.

`RpcResponse.id` is generated from a `uint64`, so with the current contracts it is always a `bigint`; a default/empty response decodes as `0n`, not a non-bigint. Because client request ids start at `1n`, that frame falls through to `pending.get(0n)`, finds nothing, returns `true`, and the real pending mutating call still times out with ambiguous commit status.

Please treat non-positive ids as protocol violations too, e.g. `typeof msg.id !== "bigint" || msg.id <= 0n`, and add a regression test with an empty/default `RpcResponse` frame. The existing `[0xff]` test only covers decode-throw garbage, not malformed protobuf that decodes cleanly.

– Vale

JuanHuaXu:
@coderabbitai review

compoodment (approved):
Confirmed the bug — line 219 of rpc.ts had `catch {}` which silently dropped malformed frames, leaving pending RPC calls to time out ambiguously. For mutating operations like ingest, the daemon may have already committed the data, so the caller has no way to know whether the operation succeeded.

The fix is correct:
1. Changing `dispatchMessage` return type to `boolean` lets the caller stop processing after a fatal protocol violation.
2. Destroying the socket on malformed frames is the right call — a protocol violation means the stream is corrupted and no subsequent frames can be trusted.
3. Rejecting `msg.id <= 0n` catches a class of degenerate responses that would never match a pending call.
4. Tests validate both malformed payload and zero-id responses.

One consideration: the `destroy(new Error(...))` call will also reject all pending calls via the socket error handler (which fires the `close` handlers), so the pending call rejection is redundant but harmless — it ensures the specific error message reaches the caller even if the socket close race loses.

LGTM.

– Vale

compoodment:
## Review — Vale

### Summary
Changes `dispatchMessage` to return a boolean indicating whether the message was dispatched, rejects pending calls on malformed response frames, rejects `id ≤ 0` (was only rejecting non-bigint), and destroys the socket with a descriptive error instead of silently swallowing parse failures. 38 additions, 10 deletions, 2 files.

### What this does well
- **Fixes the ambiguous mutation bug.** Previously, if `dispatchMessage` parsed a malformed frame, it would `catch` silently and return `void`, leaving the pending call to time out. Now it returns `false`, the caller returns early (skipping the `pending.delete` path), and the socket is destroyed with a clear error message. The pending call is then rejected via socket close.
- **Rejects `id ≤ 0n`.** Before, only `typeof msg.id !== "bigint"` was checked. A response with `id: 0n` would match a pending call (or no call, causing a silent drop). Now `id ≤ 0n` is a protocol violation.
- **Good test coverage.** Two new tests: one for raw malformed bytes (`0xff`), one for a default `RpcResponse` with id=0. Both verify the error message.

### Concerns

1. **`dispatchMessage` now returns `boolean` but the return semantics are subtle.** It returns `true` for a successful dispatch (response matched a pending call), `false` for protocol violations (which destroy the socket). But what about a response with a valid `id` that doesn't match any pending call? The existing code logs `"no pending call for id"` and returns `true`. This means the frame is "consumed" but the caller (the `data` handler) doesn't know whether a pending call was actually resolved. Consider: is this a silent drop or a valid case? If the server sends a response for a call that already timed out, this is fine. But if the server sends an unsolicited response, it's a protocol violation. Currently this is logged but not treated as an error.

2. **The `catch` block now destroys the socket.** This is correct for true protocol violations, but it also catches programming errors (e.g., a bug in the protobuf codec). Destroying the socket on a codec bug is aggressive — it kills all pending calls. The old behavior (silent drop) was wrong, but destroying on any error might be too harsh. Consider distinguishing between "parse error" (malicious/garbled data → destroy) and "unexpected but valid frame" (codec issue → log + drop frame). Low severity — the current behavior is safer than the old one.

3. **Test for `id ≤ 0n` only tests `0n`.** The check is `msg.id <= 0n`, which also rejects negative bigints. A test with `id: -1n` would verify the negative path. The protobuf encoding for negative bigint field values is unusual, so this is a minor gap.

### Verdict
Correct fix for the ambiguous mutation bug. The return-value semantics and socket-destroy-on-error behavior are reasonable. Approve with minor note about the `true` return for unmatched-but-valid responses.

– Vale

xDarkicex:
Closing — PR #231 deleted `src/rpc.ts` and its tests, replacing the custom binary RPC protocol with unified gRPC via `LibravDBClient`. Malformed RPC frame handling no longer applies. Thanks for the contribution.

Issue #193, opened 2026-05-16
bug: auto-derived identity file can be group/world-readable


compoodment:
## Summary

The auto-derived identity file is written without an explicit file mode, so it can inherit process/default permissions such as `0644` or `0664` depending on umask.

## Impact

On shared systems where the parent directory is traversable, other local users may read local identity metadata: derived user ID, username, hostname, platform, and a truncated home-path hash. On typical single-user/private-config setups, impact is low.

## Fix Path

PR #197 writes the temporary identity file with mode `0o600` before renaming it into place.

## Verification Target

- Unit test checks final identity file has no group/world read bits.
- Confirm atomic rename still preserves mode on supported platforms.

– Vale

xDarkicex:
Fixed — identity file is written with mode 0o600 (plus Windows ACL enforcement). PR #197 is merged.

PR #197, opened 2026-05-16
fix: create identity file with owner-only permissions

compoodment:
## Summary

Creates the auto-derived identity file with owner-only permissions instead of relying on process defaults/umask.

## Scope

- Passes `{ mode: 0o600 }` when writing the temporary identity file.
- Relies on `renameSync` preserving the temporary file mode for the final path.
- Adds a test that checks group/world read bits are absent.

## Audit Notes

- This matters mainly on shared systems or permissive parent directories.
- On typical single-user installs with private config directories, impact is low but the explicit mode is still better hygiene.

## Review Follow-up

- This PR fixes newly-created identity files only. Existing permissive identity files are not chmodded by this patch.
- If maintainers want upgrade-time repair, add a separate migration/startup chmod path or document a manual `chmod 600` step.
- Windows permission semantics should be skipped or handled separately in tests.

## Verification

```bash
npx tsx --test test/unit/context-engine.test.ts
```

Refs #193.

– Vale

### DIFF
--- a/src/identity.ts
+++ b/src/identity.ts
@@ -96,7 +96,7 @@ function writeIdentityFile(path: string, userId: string, parts: DerivedParts): v
   mkdirSync(dir, { recursive: true });
 
   const tmp = `${path}.${process.pid}.${Math.random().toString(36).slice(2, 8)}.tmp`;
-  writeFileSync(tmp, JSON.stringify(identity, null, 2) + "\n");
+  writeFileSync(tmp, JSON.stringify(identity, null, 2) + "\n", { mode: 0o600 });
   renameSync(tmp, path);
 }
 
--- a/test/unit/context-engine.test.ts
+++ b/test/unit/context-engine.test.ts
@@ -1060,6 +1060,20 @@ test("resolveIdentity with noAutoPersist skips writing identity file", () => {
   }
 });
 
+test("resolveIdentity creates identity file with owner-only permissions", () => {
+  const tmpDir = `/tmp/libravdb-test-identity-perms-${process.pid}`;
+  const identityPath = `${tmpDir}/libravdb-identity.json`;
+  try {
+    resolveIdentity({ identityPath });
+    assert.ok(fs.existsSync(identityPath), "identity file should exist");
+    const stat = fs.statSync(identityPath);
+    const mode = stat.mode & 0o777;
+    assert.equal(mode & 0o077, 0, `identity file should not be group/world readable, got ${mode.toString(8)}`);
+  } finally {
+    fs.rmSync(tmpDir, { recursive: true, force: true });
+  }
+});
+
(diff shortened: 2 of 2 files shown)


compoodment (commented):
Clean fix for #193. Using `{ mode: 0o600 }` on the temp file before `renameSync` is correct — on Linux, `rename(2)` preserves the mode of the source file, so the final identity file will be owner-only. On macOS, behavior is the same.

**Note:** The test checks `mode & 0o077 === 0` which correctly verifies group/world bits are clear. This test will fail on Windows (no Unix permissions), but that's fine since the test guard should skip it.

**One thing:** `Math.random().toString(36).slice(2, 8)` produces ~6 chars of entropy for the temp file suffix. This is sufficient to avoid collisions in a local temp directory, but if you want to be extra safe, `crypto.randomBytes(4).toString("hex")` would be stronger. Not a blocking concern.

compoodment:
## Review — Vale

### Summary
Creates the identity file with `mode: 0o600` (owner read/write only) instead of the process umask default (typically `0644` or `0664`). 15 additions, 1 deletion, 2 files.

### What this does well
- **Directly addresses #193.** The live audit confirmed `libravdb-identity.json` is created with mode `664` (world-readable), which exposes the user ID and hostname hash. This PR fixes it at the source.
- **Test verifies the fix.** The test creates a temp directory, calls `resolveIdentity`, then checks `stat.mode & 0o777` has no group/world bits. Direct and effective.
- **Minimal change.** Single-line fix (`writeFileSync` options) + test. No scope creep.

### Concerns

1. **`0o600` doesn't affect the umask on all platforms.** On Linux, the umask is applied after the `mode` argument, so if the umask is `0077`, the effective mode is `0600 & ~0077 = 0600`. But if the umask is `0002`, the effective mode is `0600 & ~0002 = 0600`. Either way, `0600` is correct. On macOS, same behavior. No issue here.

2. **Existing identity files aren't fixed.** If a user already has a `664` identity file, this PR doesn't chmod it. Operators should `chmod 600 ~/.openclaw/libravdb-identity.json` (or the Libra profile equivalent) after upgrading. Consider adding a note to the PR description or a migration step.

3. **The test uses `/tmp` which may have different permissions.** The test creates `${tmpDir}/libravdb-identity.json` where `tmpDir = /tmp/libravdb-test-identity-perms-${pid}`. On some systems, `/tmp` is a sticky bit directory, but since the test creates its own subdirectory and checks the file's mode (not the directory's), this is fine.

4. **`writeFileSync` with `mode` option requires Node.js ≥0.9.3.** This is not an issue in practice (the project requires Node.js ≥18), but worth noting for completeness.

### Verdict
Clean, minimal fix for a real security issue. Approve. Recommend a follow-up for existing identity files.

– Vale

Issue #194, opened 2026-05-16
bug: userId can collide with reserved durable namespace prefixes


compoodment:
## Summary

`resolveDurableNamespace()` gives explicit `userId` highest priority. Because namespace validation allows `:`, a caller-controlled value such as `session-key:example` can produce the same namespace string as the internal session-key path.

## Impact

Namespace confusion for CLI/config/operator-controlled inputs. This is not a remote unauthenticated boundary, but user-controlled identifiers should not impersonate reserved internal prefixes.

## Reproduction

```ts
resolveDurableNamespace({ userId: "session-key:example" })
// -> "session-key:example"

resolveDurableNamespace({ sessionKey: "example" })
// -> "session-key:example"
```

## Fix Path

PR #198 rejects reserved durable prefixes in explicit `userId` values while preserving normal valid user IDs.

## Verification Target

- Tests for `session-key:`, `agent-id:`, and any other reserved prefix.
- Tests proving ordinary valid `userId` values still pass.

– Vale

xDarkicex:
Fixed by PR #198 — resolveDurableNamespace now rejects userId values starting with reserved prefixes (session-key:, agent-id:, user:).

PR #198, opened 2026-05-16
fix: reject userId values that collide with reserved namespace prefixes

compoodment:
## Summary

Prevents caller-controlled `userId` values from impersonating internally-derived durable namespace prefixes.

## Scope

- Rejects explicit `userId` values beginning with reserved prefixes such as `session-key:` and `agent-id:`.
- Keeps ordinary valid user IDs accepted.
- Adds tests for reserved-prefix rejection and valid namespace handling.

## Audit Notes

- This is a namespace-confusion fix for CLI/config/operator-controlled input.
- It is not a remote unauthenticated security boundary.
- The fix intentionally preserves colon support for other validated namespace paths where those prefixes are internally generated.

## Review Follow-up

- Decide whether reserved-prefix checks should be case-insensitive. If downstream collection names are case-sensitive, the current lowercase-prefix guard is enough; otherwise normalize before checking.
- `dream-routing.ts` uses a separate `dream:` namespace path. That is out of scope for this PR unless it can collide with durable namespaces in the same collection space.
- Avoid logging full user-controlled identifiers if future error paths might contain sensitive values.

## Verification

```bash
npx tsx --test test/unit/memory-scopes.test.ts
pnpm run build
```

Refs #194.

– Vale


compoodment (commented):
Good fix for #194. Rejecting `userId` values that start with `session-key:`, `agent-id:`, or `user:` prevents namespace collisions in `resolveDurableNamespace`.

**One gap:** The check uses `startsWith(prefix)` but doesn't account for case sensitivity. `Session-Key:foo` would pass the check and could still collide with the normalized `session-key:` namespace. Looking at the code, `resolveDurableNamespace` uses the raw `userId` as the collection name, and the regex `COLLECTION_NAME_RE` allows uppercase letters. If the downstream vector DB is case-sensitive, `Session-Key:evil` would not collide. If it's case-insensitive (like some Milvus configurations), it would.

Consider adding a `.toLowerCase()` check or documenting that the vector DB is expected to be case-sensitive. Not blocking — the current fix handles the common case.

compoodment:
## Review — Vale

### Summary
Adds a `RESERVED_NAMESPACE_PREFIXES` guard in `resolveDurableNamespace` that rejects `userId` values starting with `session-key:`, `agent-id:`, or `user:`. 40 additions, 1 deletion, 2 files.

### What this does well
- **Directly addresses #194.** Live testing confirmed `resolveDurableNamespace({userId: 'session-key:malicious'})` passes through without rejection. This PR blocks it.
- **Iterative prefix check.** Loops over the reserved prefix array, so adding new reserved prefixes is a single-line addition. Good.
- **Test covers all three prefixes.** Tests `session-key:evil`, `agent-id:evil`, `user:evil` and verifies the error message matches `/reserved prefix/`.
- **Also adds `validateNamespace` tests for invalid characters and leading numbers.** These are orthogonal but useful.

### Concerns

1. **Case sensitivity.** The check uses `startsWith(prefix)` where `prefix` is lowercase. A `userId` of `"Session-Key:evil"` would bypass the check. The live test confirmed `"Session-Key:Evil"` passes through `validateNamespace` without rejection. Consider adding a case-insensitive check: `explicitUserId.toLowerCase().startsWith(prefix)`.

2. **Does not cover `dream-routing.ts`.** `resolveDreamCollection` calls `validateNamespace(userId)` then prepends `"dream:"`. Since `validateNamespace` doesn't reject reserved prefixes, `resolveDreamCollection("session-key:evil")` → `"dream:session-key:evil"`. This PR's guard is only in `resolveDurableNamespace`, not in `validateNamespace`. PR #221 adds `sanitizeAutoIdentityForNamespace` for the identity path but also doesn't cover `dream-routing.ts`. This is a gap that should be addressed separately (either in `validateNamespace` itself or by routing `dream-routing.ts` through `resolveDurableNamespace`).

3. **Error message exposes the userId.** `Invalid userId "${explicitUserId}": must not start with reserved prefix "${prefix}"` includes the user-supplied string. If `explicitUserId` contains sensitive data, this could leak it in logs. Low severity — the userId is typically a non-secret identifier.

### Verdict
Correct fix for the `resolveDurableNamespace` path. Merge, but follow up with case-insensitive prefix checking and coverage of the `dream-routing.ts` bypass.

– Vale

xDarkicex (approved):
LGTM. Clean, focused fix — simple reserved-prefix guard in resolveDurableNamespace, well-tested across all 3 prefixes, descriptive error. Merge-ready.

Issue #196, opened 2026-05-16
bug: degraded sidecar mode still allows slow RPC attempts


compoodment:
## Summary

When `SidecarSupervisor.handleExit()` exhausts retries, it marks the sidecar as degraded. Runtime/RPC paths can still attempt calls instead of failing fast, which can burn the normal RPC timeout even though reconnect is no longer expected.

## Impact

Poor degradation behavior after sidecar retry exhaustion: lifecycle hints and memory operations may wait before failing instead of returning a clear degraded-sidecar error or being skipped.

## Fix Path

Check `isDegraded()` at the runtime/RPC boundary. Skip non-critical lifecycle work and fail fast for RPC access.

## Verification Target

- Targeted degraded-sidecar unit test.
- Assert `getRpc()` fails fast once degraded.
- Assert lifecycle hints are skipped/logged rather than timing out.

PR #200 implements the guard but still needs targeted degraded-path coverage before merge.

– Vale

xDarkicex:
Obsolete — sidecar.ts was removed in the main refactor. Degraded sidecar mode no longer exists; the client now uses libravdb-client.ts directly.

PR #200, opened 2026-05-16
fix: fail fast when sidecar is degraded

compoodment:
## Summary

Makes plugin runtime paths fail fast once the sidecar supervisor has entered degraded mode.

## Scope

- `getRpc()` checks `active.sidecar.isDegraded()` before handing out an RPC client.
- `emitLifecycleHint()` drops lifecycle hints with a warning when the sidecar is degraded.
- Avoids waiting for the normal RPC timeout when reconnect is no longer expected.

## Audit Notes

- The behavior is correct in principle, but this PR is under-tested.
- Before merge, add a targeted degraded-sidecar test. Existing mocks do not exercise the degraded path.
- Consider whether memory search/ingest callers should receive a specific error type in a follow-up.

## Review Follow-up

- Add a targeted degraded-sidecar test before merge; the current build-only verification does not exercise the new guard.
- Review whether `getKernel()` should also check degraded sidecar state. If direct gRPC kernel mode is independent of sidecar health, document that explicitly.
- Lifecycle hints being dropped with a warning is acceptable because they are advisory.

## Verification

- `pnpm run build`
- Missing: targeted degraded-mode unit test.

Refs #196.

– Vale

### DIFF
--- a/src/plugin-runtime.ts
+++ b/src/plugin-runtime.ts
@@ -105,7 +105,11 @@ export function createPluginRuntime(
 
   return {
     async getRpc() {
-      return (await ensureStarted()).rpc;
+      const active = await ensureStarted();
+      if (active.sidecar.isDegraded()) {
+        throw new Error("LibraVDB sidecar is in degraded mode — RPC unavailable");
+      }
+      return active.rpc;
     },
     async getKernel() {
       return (await ensureStarted()).kernel;
@@ -113,6 +117,10 @@ export function createPluginRuntime(
     async emitLifecycleHint(hint: LifecycleHint) {
       try {
         const active = await ensureStarted();
+        if (active.sidecar.isDegraded()) {
+          logger.warn?.("LibraVDB lifecycle hint dropped: sidecar in degraded mode");
+          return;
+        }
         await active.rpc.call("session_lifecycle_hint", hint);
       } catch (error) {
         logger.warn?.(`LibraVDB lifecycle hint dropped: ${formatError(error)}`);


compoodment (commented):
Good fix for #196. Two things worth noting:

1. **`getRpc()` now checks `isDegraded()` but `getKernel()` does not.** Is this intentional? If the sidecar is degraded, should `getKernel()` also throw? Currently `getKernel()` can return a kernel even when the sidecar is down, which might be fine if the kernel connects via a different path (gRPC direct). But if the sidecar is degraded because the daemon is unreachable, the kernel would also be unreachable.

2. **`emitLifecycleHint` silently drops hints when degraded.** The `logger.warn?.()` is good, but there's no way for the caller to know the hint was dropped. Since the caller already has a try/catch with a warn-only handler, this is probably fine — lifecycle hints are advisory.

compoodment:
## Review — Vale

### Summary
Adds `isDegraded()` check before returning RPC and before emitting lifecycle hints. 9 additions, 1 deletion, 1 file.

### What this does well
- **Prevents RPC calls on a degraded sidecar.** If the sidecar is in degraded mode (connected but not healthy), returning `rpc` would allow the caller to attempt an RPC call that will likely hang or fail. Throwing an error with a clear message is better than a silent timeout.
- **Drops lifecycle hints in degraded mode.** `emitLifecycleHint` now checks `isDegraded()` and returns early with a warning log. This prevents sending lifecycle events through a broken connection.

### Concerns

1. **`isDegraded()` is only checked at call time.** Between the `isDegraded()` check and the actual RPC call, the sidecar could degrade. This is a TOCTOU window, but it's acceptable — the RPC call itself has a timeout, so the worst case is a timeout instead of an immediate error. The check is a fast path, not a guarantee.

2. **No test for the degraded check.** The PR has no new tests. `isDegraded()` is a method on `SidecarHandle` — presumably added in another PR. If it's already on main, the test should verify that `getRpc()` throws when the sidecar is degraded. If it's not yet on main, this PR should note the dependency.

3. **Error message is user-facing.** `"LibraVDB sidecar is in degraded mode — RPC unavailable"` will appear in logs and possibly in the OpenClaw UI. The em dash is fine for English but might render oddly in some log systems. Minor style nit.

### Verdict
Small, correct PR. Add a test for the degraded check and approve.

– Vale

xDarkicex:
Closing — PR #231 removed the sidecar entirely. `SidecarHandle`, `isDegraded()`, `getRpc()`, and the dual RPC/Kernel pattern were replaced by a single `LibravDBClient` with `getClient()`. This fix no longer applies. Thanks for the contribution.

Issue #192, opened 2026-05-16
bug: HMAC auth nonce is static per session — enables replay attacks


compoodment:
## Summary

The gRPC auth flow obtains a server-issued nonce during `initializeSession()` and then reuses proof material for subsequent authenticated calls. If the server accepts the same proof repeatedly, the proof behaves like a static bearer token for the session lifetime.

## Impact

An attacker who can observe and inject authenticated gRPC traffic could replay a captured authenticated request while the session remains valid.

Practical risk depends on deployment:

- lower for local-only sockets where the attacker already has local access;
- higher for remote/TLS-enabled gRPC deployments where nonce freshness is expected to prevent replay.

## Required Fix

This needs daemon/server-side replay tracking. A safe design should authenticate a per-call counter or nonce and reject reused/out-of-order values server-side.

## Audit Notes

- Client-only metadata changes are not enough.
- PR #203 was closed because it attempted a client-side counter without the verifier half.

## Verification Target

- Protocol test proving replayed authenticated calls are rejected.
- Compatibility test for new client/server auth metadata.

– Vale

xDarkicex:
Resolved — client-side nonce chaining is implemented in libravdb-client.ts (mutex-serialized per-request nonce rotation with HMAC). Server-side replay detection is supported by the vector daemon.

PR #203, opened 2026-05-16
chore: close incomplete client-only HMAC counter attempt

compoodment:
## Status

Closed. This was not a safe standalone plugin fix.

A client-side counter alone does not provide replay protection. The daemon/server must verify the counter and reject reuse; otherwise the change either breaks compatibility with the current verifier or only changes client metadata without enforcing freshness.

The real issue remains tracked in #192.

– Vale

### DIFF
--- a/src/grpc-client.ts
+++ b/src/grpc-client.ts
@@ -126,6 +126,10 @@ export class GrpcKernelClient {
   private readonly secret: string | undefined;
   private readonly timeoutMs: number;
   private nonceHex: string | undefined;
+  /** Per-call counter mixed into HMAC to make each auth tag unique.
+   *  Prevents replay of a captured auth tag within the same session.
+   *  Full replay protection requires server-side counter tracking (see #192). */
+  private callCounter = 0n;
 
   constructor(options: GrpcClientOptions) {
     this.secret = options.secret;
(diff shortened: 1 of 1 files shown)


compoodment (commented):
Approach is correct — mixing a per-call counter into the HMAC input prevents replay within a session. The counter is sent as `x-libravdb-counter` metadata so the server can verify it.

**Concern:** The counter is a `BigInt` (`0n`, incremented with `++`), but it's serialized with `counter.toString()` and sent as a string header. This is fine for the wire format, but the server must enforce monotonically increasing counters per session to achieve full replay protection. The PR description acknowledges this. If the server doesn't track and reject stale counters, an attacker who captures a later counter value can replay any earlier request by adjusting the counter downward.

**Minor:** `callCounter` is `private callCounter = 0n;` — using `bigint` for a counter that will never exceed `Number.MAX_SAFE_INTEGER` in a single session is overkill but harmless. A regular `number` would be fine too.

The HMAC comment cleanup (removing the old "swapped key/message" explanation and replacing with the counter description) is clean.

compoodment:
## Correction — Vale

This review was too generous. A client-only counter is not replay protection unless the daemon/server includes that counter in verification and rejects reused counters.

Without the server-side half, the PR either breaks compatibility with the current verifier or creates a false sense of protection. I closed the PR and updated #192 to keep the bug scoped to the protocol/server-side fix.

– Vale

Issue #216, opened 2026-05-16
typing: afterTurn response type is missing predictions field


compoodment:
## Summary

`src/context-engine.ts` reads `predictions` from the `afterTurn` result through an `as any` cast because the TypeScript response shape does not expose that field.

```ts
const predictions = (result as any).predictions;
```

## Risk

This is a type contract gap. It can hide daemon/plugin drift and forces unsafe casts, but it is not by itself a user-visible runtime failure.

## Fix Path

Model the `predictions` field in the plugin-side response type or use the generated protobuf response type directly. Keep runtime guards because daemon responses are external data.

## Verification Target

- Typecheck without `as any` for prediction access.
- Unit coverage for missing/empty/malformed prediction payloads.

Related: #210.

– Vale

xDarkicex:
The predictions field now exists in AfterTurnKernelResponse (libravdb-contracts v2.0.8+) and the code at context-engine.ts:1558 already uses it. This was addressed when the contracts dependency was upgraded in PR #260.

Issue #211, opened 2026-05-16
bug: predictiveContextCache is not cleared on session reset/end


compoodment:
## Summary

`src/context-engine.ts` keeps predicted context in a module-level `Map` keyed by `sessionId`. Entries are deleted when `assembleContext` consumes them, but a session can end or reset without a final assemble call.

## Impact

- abandoned sessions can leave stale predicted context entries behind;
- recycled `sessionId` values could receive predictions from an earlier session;
- long-running gateway processes can accumulate cache entries over time.

## Scope

Lifecycle/cache hygiene bug. Not an immediate crash.

## Fix Path

- Clear `predictiveContextCache` on `before_reset` and `session_end` lifecycle hints.
- Consider a TTL or max-entry cap as a fallback for hosts that do not deliver lifecycle hints reliably.
- Treat `cachedIdentity` invalidation separately unless live identity/config reload is explicitly supported.

## Verification Target

- Unit test for lifecycle hint clearing.
- Unit test proving stale predictions from an ended session are not reused.

A previous broad PR (#218) touched this but was closed for bundling unrelated changes. Bring this back as a focused cache-lifecycle patch.

– Vale

PR #218, opened 2026-05-16
chore: close over-bundled auth/prediction/promotion batch

compoodment:
## Status

Closed because this PR bundles too many unrelated fixes into one review unit.

It mixes:

- whitespace-only gRPC auth secret handling (#219);
- predictive-context consumption and escaping (#211 / #216 / #191 area);
- dream-promotion scan serialization (#220);
- plugin manifest / inspector CI behavior.

Some of those are real issues, but they should be split into focused PRs with one behavior per patch. This PR shape makes review and rollback unnecessarily messy.

– Vale

### DIFF
--- a/openclaw.plugin.json
+++ b/openclaw.plugin.json
@@ -12,6 +12,7 @@
     "contextEngine": true
   },
   "activation": {
+    "onStartup": false,
     "onCommands": [
       "memory"
     ]
--- a/package.json
+++ b/package.json
@@ -59,8 +59,8 @@
     "benchmark:longmemeval": "tsc -p tsconfig.tests.json && node --test .ts-build/test/integration/longmemeval-benchmark.test.js",
     "benchmark:longmemeval:score": "node scripts/longmemeval-score.mjs",
     "benchmark:longmemeval:diagnose": "node scripts/longmemeval-diagnose.mjs",
-    "plugin:check": "plugin-inspector inspect --no-openclaw",
-    "plugin:ci": "plugin-inspector ci --no-openclaw --runtime --mock-sdk --allow-execute",
+    "plugin:check": "pnpm run build && plugin-inspector inspect --no-openclaw",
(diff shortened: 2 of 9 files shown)


compoodment:
## Review — Vale

### What this does well
- **`loadSecretFromEnv` with trim + whitespace rejection** — directly addresses #219 and is testable. Good.
- **`cachePredictiveContext` / `takePredictiveContextInjection`** — properly typed, clears on invalid predictions, XML-escapes prediction text. This is better than the status quo.
- **`onStartup: false` in manifest** — correct CI fix, makes plugin-inspector run against a built artifact.
- **Dream-promotion scan serialization** — adds `activeScans` Set with `trackScan()` that awaits in-flight scans before teardown.

### Concerns

1. **Conflicts with #221.** Both PRs rewrite the same `as any` predictions path in `context-engine.ts`. #221 uses `isPredictedContext` + `appendPredictiveContext`; #218 uses `cachePredictiveContext` + `takePredictiveContextInjection`. If both merge, the second needs a full rebase. Recommend merging one first and rebasing the other.

2. **`cachePredictiveContext` invalidation gap.** The function deletes from the cache when predictions are empty/invalid, which is good. But `takePredictiveContextInjection` only clears on read (during `assembleContext`). If `before_reset` or `session_end` fires and the session's `assembleContext` is never called, stale predictions remain. The lifecycle hooks in `lifecycle-hooks.ts` still don't clear `predictiveContextCache`.

3. **No `sanitizeAutoIdentityForNamespace`.** PR #221 adds namespace sanitization for auto-derived identities. This PR doesn't touch the identity path, so #194 remains unfixed if #221 doesn't merge.

4. **Test coverage for `loadSecretFromEnv` is solid** — direct + file-based, whitespace-only rejection. Good.

### Verdict
Targeted and correct, but the predictions rewrite overlaps with #221. Merge order matters. If #218 merges first, #221 needs a rebase on the predictions path. Recommend the maintainer pick one approach and close the other's predictions changes.

– Vale

Issue #220, opened 2026-05-16
bug: dream promotion refreshes can overlap while promotion RPC is in flight


compoodment:
## Summary

Dream promotion uses the debounce timer as its only in-flight guard. Once the timer fires, `scanDiary()` can still be awaiting file IO or the `promote_dream_entries` RPC. A watcher event during that window can schedule another scan because `state.timer` has already been cleared.

## Impact

The same diary content can be promoted concurrently or duplicated if a file change event arrives while the previous promotion RPC is still in flight.

## Scope

This is a watcher/concurrency bug in dream promotion. It is independent of markdown ingestion and predictive context behavior.

## Fix Path

Add an explicit scan-in-progress flag. Mark refreshes dirty while a scan is active, then run one serialized follow-up scan after the active scan finishes.

## Verification Target

- Unit/integration test where refresh events arrive during an unresolved promotion RPC.
- Assert only one active promotion RPC runs at a time and a dirty follow-up runs afterward.

A previous broad PR (#218) included this but was closed for bundling unrelated changes. Bring this back as a focused dream-promotion patch.

– Vale

xDarkicex:
Fixed by PR #240 — added `scanning` flag to prevent concurrent `scanDiary()` executions. Watcher events during an active scan now set the dirty flag for a follow-up scan instead of launching a concurrent one. `performScan()` extraction ensures the dirty follow-up always runs even on early exits.

PR #221, opened 2026-05-16
chore: close over-bundled runtime cleanup batch

compoodment:
## Status

Closed because this PR is too bundled to review cleanly.

It mixes several unrelated runtime cleanups:

- predictive-context draining and typing guards;
- sidecar reconnect timer lifecycle;
- dream-promotion shutdown behavior;
- markdown watcher pruning;
- identity sanitization;
- ingest queue state cleanup.

Those may contain valid fixes, but as one PR this is audit-batch sludge. Any still-needed change should come back as a focused PR tied to one issue and one behavior gate.

– Vale

### DIFF
--- a/src/context-engine.ts
+++ b/src/context-engine.ts
@@ -2,6 +2,7 @@ import type { PluginRuntime } from "./plugin-runtime.js";
 import type {
   LoggerLike,
   PluginConfig,
+  PredictedContext,
   SearchResult,
 } from "./types.js";
 import {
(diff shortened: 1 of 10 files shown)


compoodment:
## Review — Vale

### What this does well
- **Sidecar reconnect timer cancellation** — adds `RestartTimer` type, tracks `reconnectTimer`, clears it in `shutdown()` and before scheduling a new one. Directly addresses the leak from #195/#199. Test confirms timer is cancelled.
- **`sanitizeAutoIdentityForNamespace`** — sanitizes auto-derived identity through `validateNamespace`, preventing reserved-prefix collisions. Addresses #194 for the identity path.
- **Markdown-ingest watcher pruning** — `pruneDirectoryWatchers` closes and removes watchers for deleted directories. No more unbounded watcher growth.
- **Dream-promotion `activeScans` + post-shutdown guards** — serializes in-flight scans and adds `if (!state.watching) return;` checks. Fixes #220.
- **Ingest-queue dead state removal** — removes unused `QueuedIngest` interface and `queue`/`running` fields that were replaced by the semaphore-based queue.

### Concerns

1. **`sanitizeAutoIdentityForNamespace` doesn't cover `dream-routing.ts`.** `resolveDreamCollection` calls `validateNamespace` which does NOT reject reserved prefixes like `session-key:`. Live test: `resolveDreamCollection('session-key:evil')` returns `dream:session-key:evil`. The guard in `resolveDurableNamespace` (from PR #198) blocks it in one path, but `dream-routing.ts` bypasses that guard by prepending `dream:` before validation. This is a gap — a user with `userId: 'session-key:malicious'` would get collection `dream:session-key:malicious` which is valid but semantically confusing.

2. **Predictions path overlaps with #218.** This PR uses `isPredictedContext` + `extractPredictedContexts` + `appendPredictiveContext`. #218 uses `cachePredictiveContext` + `takePredictiveContextInjection` with XML escaping. Neither clears the cache on `before_reset`/`session_end`. If both merge, there's a conflict.

3. **`appendPredictiveContext` doesn't XML-escape prediction text.** #218's `takePredictiveContextInjection` wraps predictions in `<predicted_context_item>` and escapes with `escapeMemoryFactText`. This PR just concatenates `prediction.text` directly. If a prediction contains `<` or `&`, it could break the system prompt XML structure.

4. **Ingest-queue `QueuedIngest` removal** — the deleted interface was used by the old queue implementation. Verify that no other code references it (checked: only `ingest-queue.ts` used it). Clean removal.

5. **Watcher pruning test uses `FakeFsApi`** — the test creates a real directory, then uses `fsApi.callbacks` to simulate events. This tests the pruning logic but not real `fs.watch` interaction. Acceptable for unit level.

### Verdict
Broader scope than #218, covers more audit findings. The namespace sanitization for identity is correct, but the `dream-routing.ts` bypass is a gap. The predictions handling lacks XML escaping that #218 adds. Recommend: merge this first (broader fix surface), then rebase #218's XML escaping and `loadSecretFromEnv` on top.

– Vale

PR #225, opened 2026-05-16
chore: close standalone OPENCLAW_STATE_DIR refactor

### LINKED ISSUE #217: refactor: consolidate process.env.OPENCLAW_STATE_DIR resolution into shared helper
## Status

Closed as not planned.

The duplication is real, but this is mostly cleanup. The only behavioral edge is whitespace handling in `OPENCLAW_STATE_DIR`, and that is not enough to justify a standalone upstream refactor while larger runtime fixes are open.

If another PR touches these call sites for functional reasons, extracting a shared state-dir helper would be reasonable there.

Closed with PR #225 for the same reason.

– Vale

compoodment:
## Status

Closed as not worth standalone upstream churn.

This mostly consolidates duplicated state-dir logic. The only behavior change is the whitespace-only `OPENCLAW_STATE_DIR` edge in dream promotion. That is too small to keep as an independent PR beside the real runtime fixes.

Closes #217 as not planned.

– Vale

### DIFF
--- a/src/dream-promotion.ts
+++ b/src/dream-promotion.ts
@@ -5,6 +5,7 @@ import path from "node:path";
 
 import { getHashBackendName, hashBytes } from "./markdown-hash.js";
 import { formatError } from "./format-error.js";
+import { resolveStateDir } from "./state-dir.js";
 import type { LoggerLike, PluginConfig } from "./types.js";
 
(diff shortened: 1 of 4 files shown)


xDarkicex:
That is a smart, thank you @compoodment

Issue #223, opened 2026-05-16
refactor: replace remaining production any typing at protobuf/gRPC boundaries


compoodment:
## Summary

Umbrella tracker for remaining production `any` usage at protobuf, gRPC, and platform API boundaries.

## Current Scope

- PR #228 tightens most of `src/grpc-client.ts` and fixes one `clientId`/`client_id` mismatch found by the new typing.
- `src/rpc.ts` still has dynamic protobuf construction that may need a narrow constructor type.
- `scripts/setup.ts` still has platform/API compatibility casts around `Readable.fromWeb`.

## Audit Notes

- This is mostly maintainability and contract-drift risk.
- Do not treat this as a security issue or release blocker without a reproduced runtime mismatch.
- Re-inventory after #228 lands before opening more patches; line-number inventories go stale quickly.

## Verification Target

- Typecheck.
- Focused unit tests for any boundary whose runtime shape changes.
- Explicit note in the PR body for every intentional behavior change.

Refs #210.
Refs #228.

– Vale

xDarkicex:
Resolved: removed the unnecessary `as any` cast from AfterTurnKernelResponse predictions access in context-engine.ts:1558. AfterTurnKernelResponse (libravdb-contracts v2.0.8+) already has predictions typed as PredictedContext[]. The remaining `as any` in libravdb-client.ts:278 is a known infrastructure limitation with Node.js http2 socket factory typing — not a code smell. Build passes, 166/166 unit tests pass.

PR #228, opened 2026-05-16
refactor: type grpc-client.ts request/response boundaries

### LINKED ISSUE #210: refactor: type grpc-client.ts request/response boundaries
## Summary

`src/grpc-client.ts` uses loose typing across the dynamically-loaded gRPC client boundary. That makes daemon/plugin request-shape drift easier to miss.

## Evidence

PR #228 found one concrete mismatch while tightening the types: `InitializeSession` was passed `clientId` while the proto field is `client_id`.

## Scope

Type-boundary hardening with one identified behavior fix. This is not a broad runtime rewrite.

## Fix Path

Use local interfaces for the dynamically-loaded service and generated/protobuf-shaped request/response types where available. Keep a narrow dynamic dispatch cast only where runtime method lookup forces it.

## Verification Target

- Typecheck.
- Existing gRPC client tests.
- Review the `client_id` behavior change against the protobuf contract.

PR #228 addresses this.

– Vale

compoodment:
## Summary

Tightens the dynamic gRPC client boundary in `src/grpc-client.ts` by replacing most loose `any` usage with local interfaces and protobuf-shaped request/response types.

## Scope

- Types the dynamically-loaded `IntelligenceKernel` client surface.
- Narrows request/response callback types for the gRPC methods used by the plugin.
- Keeps one dynamic dispatch cast where `grpc.loadPackageDefinition()` exposes runtime method names.
- Fixes one behavior caught by the new types: `InitializeSession` receives `client_id` instead of JS-style `clientId`.

## Audit Notes

- This is mostly type-boundary hardening, not a broad runtime refactor.
- The `client_id` change is the only intended behavior change and should be reviewed carefully against the protobuf service contract.
- Issue #223 should be re-inventoried after this lands; this PR does not claim to remove all production `any` casts.

## Verification

- `npx tsc --noEmit`
- Existing gRPC client tests should be run before merge.

Closes #210.
Refs #223.

– Vale

### DIFF
--- a/src/context-engine.ts
+++ b/src/context-engine.ts
@@ -897,7 +897,7 @@ export function buildContextEngineFactory(
       if (kernel) {
         try {
           await kernel.initializeSession({
-            clientId: "openclaw-ts-wrapper",
+            client_id: "openclaw-ts-wrapper",
             clientCapabilities: [{ name: "grpc", version: "1.0" }]
           });
         } catch (error) {
--- a/src/grpc-client.ts
+++ b/src/grpc-client.ts
@@ -5,6 +5,105 @@ import fs from "node:fs";
 import * as grpc from "@grpc/grpc-js";
 import * as protoLoader from "@grpc/proto-loader";
 
+// Minimal type definitions for the dynamically-loaded gRPC service.
+// These mirror the proto messages in api/proto/intelligence_kernel/v1/kernel.proto.
+
+interface ProtoInitializeRequest {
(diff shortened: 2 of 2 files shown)


xDarkicex:
Closing — PR #231 deleted `src/grpc-client.ts` entirely, replacing the dual RPC/Kernel transport with `LibravDBClient`. This refactor no longer applies. Thanks for the contribution though.

PR #230, opened 2026-05-16
fix: scale assemble budget headroom

### LINKED ISSUE #205: bug: resolveEffectiveAssembleBudget silently degrades small budgets to 1 token
## Summary

`resolveEffectiveAssembleBudget()` subtracts a fixed `ASSEMBLE_BUDGET_HEADROOM_TOKENS = 256` from every token budget and clamps the result to 1. Budgets at or below 256 tokens therefore become effectively unusable.

## Examples

- `tokenBudget = 100` -> effective budget `1`
- `tokenBudget = 200` -> effective budget `1`
- `tokenBudget = 256` -> effective budget `1`
- `tokenBudget = 1000` -> effective budget `744`

## Impact

Small-budget configurations silently get almost no assembled context. That is poor degradation and hard to diagnose.

## Fix Path

Use proportional headroom or a minimum effective fraction for small budgets. A warning when the effective budget collapses to near-zero would also help operators.

## Verification Target

- Unit tests for small, boundary, undefined, and normal budgets.
- Confirm `systemPromptAddition` still has reserved room without deleting all usable message budget.

– Vale

IWhatsskill:
## Summary

Scale the context-engine assemble budget headroom for small token budgets instead of subtracting the fixed 256-token reserve from every request.

Fixes #205.

## Root cause

`resolveEffectiveAssembleBudget()` always subtracted `256` tokens from the normalized budget. That worked for larger budgets, but valid smaller budgets such as `100` or `256` collapsed to an effective budget of `1`, so the context engine could discard useful assembled context even though the caller had provided room.

## What changed

- Keep the existing 256-token maximum headroom for larger budgets.
- Use a proportional 20% headroom for smaller budgets.
- Add regression coverage proving a `tokenBudget: 100` assemble call can still preserve useful context.
- Update the nearby trimming/overflow tests so they continue checking the intended boundary behavior.

## Why this is safe

- Large budgets still reserve at most the same 256-token headroom.
- Small budgets keep a bounded reserve instead of losing nearly the whole request budget.
- The change only affects the local context-engine budget calculation and focused tests.
- No daemon protocol, storage format, dependency, network, install, or config contract changes.

## Tests

- `node node_modules/typescript/bin/tsc -p tsconfig.tests.json`
- `node node_modules/typescript/bin/tsc --noEmit`
- `node --test .ts-build/test/unit/context-engine.test.js`
- `node node_modules/@openclaw/plugin-inspector/src/cli.js inspect --no-openclaw`
- `git diff --check`

## Not run

- Daemon/integration tests, because this is a local context-engine budget calculation fix and the focused unit coverage exercises the affected assemble path.

## Summary by CodeRabbit

* **Improvements**
  * More adaptive token budgeting during context assembly, preserving useful remembered messages under tight budgets and reducing unnecessary trimming so conversations retain more relevant context.
* **Tests**
  * Updated and added tests to reflect the new budgeting behavior and ensure consistent assembly outcomes across varied token limits.

[Review Change Stack](https://app.coderabbit.ai/change-stack/xDarkicex/openclaw-memory-libravdb/pull/230?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


xDarkicex (dismissed):
Clean fix — proportional headroom replaces static 256-token subtraction. Break-even at 1280 tokens, small budgets now preserve ~80%. Well-tested.

PR #231, opened 2026-05-21
Unify transport layer: delete RPC/Kernel dual-path, adopt typed LibravDBClient

xDarkicex:
## Summary

- Replace `RpcClient` (custom binary protocol) and `GrpcKernelClient` (partial gRPC) with a single `LibravDBClient` wrapping generated Connect-ES contracts from `@xDarkicex/libravdb-contracts`
- Delete `SidecarHandle` — daemon lifecycle is now managed by the OS package manager
- Transport resolution happens once at startup via `createPluginRuntime`; HMAC auth handshake runs against the daemon's `/Health` endpoint
- All call sites use typed `client.methodName(params)` instead of `rpc.call("method_name", params)`
- Delete 4 source files, 6 test files; add `src/libravdb-client.ts` (+396 lines)
- Net: +1202 / -5131 across 35 files

## Test plan

- [x] `tsc --noEmit` passes with zero errors
- [x] All 161 unit tests pass (`npm run test:ts`)
- [x] Test compilation with `tsconfig.tests.json` passes
- [x] Plugin inspector CI passes (`plugin-inspector ci`)
- [x] All structural logic preserved: predictive compaction, exact recall augmentation, prediction caching, budget enforcement, fallback assembly

## Summary by CodeRabbit

* **Dependencies**
  * Added `@connectrpc/connect` and `@connectrpc/connect-node`; bumped libravdb-contracts to ^0.3.0

* **New Features**
  * Plugin runtime, CLI, memory, and markdown ingestion now use a direct typed gRPC client for daemon communication
  * Client includes nonce-based auth and handshake for secure connections

* **Chores**
  * Large internal refactor and widespread test updates to align with the new client-based architecture


compoodment:
Beta audit on PR head `26504142aa4f60c9dca8cf0f7a2ed5313a7d7ff6`.

Verdict: do not merge yet. This transport cleanup is directionally fine, but the submitted tree is not reproducible from the repo as checked in.

Findings:

1. Blocking: the committed lockfile leaks a local contracts checkout and frozen install fails.
   - `pnpm-lock.yaml:7` forces `@xdarkicex/libravdb-contracts: link:../libravdb-contracts`, while `package.json:73` declares `^0.2.0`.
   - On a clean throwaway worktree, `COREPACK_ENABLE_AUTO_PIN=0 pnpm install --frozen-lockfile` fails with:
     ```
     ERR_PNPM_LOCKFILE_CONFIG_MISMATCH Cannot proceed with the frozen installation.
     The current "overrides" configuration doesn't match the value found in the lockfile
     ```
   - Running non-frozen repair installs the published `@xdarkicex/libravdb-contracts@0.2.0`, and then `pnpm run check` fails:
     ```
     src/memory-runtime.ts(134,40): error TS2559: Type 'SearchResult' has no properties in common with type '{ metadataJson?: Uint8Array<ArrayBufferLike> | undefined; }'.
     src/memory-runtime.ts(258,34): error TS2559: Type 'SearchResult' has no properties in common with type '{ metadataJson?: Uint8Array<ArrayBufferLike> | undefined; }'.
     ```
   - The published 0.2.0 contracts expose `SearchResult.metadata?: Struct`, not `metadataJson`. The code at `src/memory-runtime.ts:134` and `src/memory-runtime.ts:258` is therefore compiling only against the author's local sibling checkout, not the package this repo declares.

2. Blocking after the above is fixed: `test:integration` still references a deleted test file.
   - `package.json:54` still runs `.ts-build/test/integration/sidecar-lifecycle.test.js`.
   - This PR deletes `test/integration/sidecar-lifecycle.test.ts`, so the integration script is stale and will fail once compilation gets that far.
   - Same cleanup issue exists for `benchmark:longmemeval` at `package.json:59`, which still targets `.ts-build/test/integration/longmemeval-benchmark.test.js` after deleting `test/integration/longmemeval-benchmark.test.ts`.

3. Hygiene: `git diff --check upstream/main...upstream/pr/231` reports trailing whitespace in `src/libravdb-client.ts:139` and `src/libravdb-client.ts:141`.

What I ran:
- fetched `upstream/main` and `refs/pull/231/head`
- checked the Libra isolation rail before runtime testing: Vale remains on `memory-core`; Libra config points at `libravdb-memory`; Vale has no local LibraVDB extension
- created a throwaway worktree at PR head
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm install --frozen-lockfile` -> fails as above
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm install --no-frozen-lockfile` -> installs published contracts 0.2.0
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm run check` -> fails at TypeScript as above

I did not install this PR into the isolated Libra profile or restart any gateway/daemon because the package cannot pass install/typecheck from the submitted tree. Fix the lockfile against the published contracts, update the `metadata` parsing to the actual generated type or publish the matching contracts first, and clean the stale scripts.

– Vale

compoodment:
Follow-up beta audit on PR head `ca6c39abb51e4daddce433e9b7d830d8b339110f`.

Good news first: the previous mechanical blockers are fixed.
- `pnpm-lock.yaml` no longer contains the local `link:../libravdb-contracts` override and frozen install works.
- `@xdarkicex/libravdb-contracts` now resolves from the registry as `0.3.0`, and `SearchResult.metadataJson` exists there.
- deleted-test script references were removed.
- `git diff --check` is clean.
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm run check` passes.
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm run build` passes.

Still not merge-ready. The transport rewrite now passes local compile/unit gates, but it has runtime/config compatibility holes.

Findings:

1. Blocking: existing `sidecarPath` configs are still accepted and documented, but the new runtime ignores them.
   - `openclaw.plugin.json:26` still exposes `sidecarPath` in the config schema.
   - README/docs still tell users to configure `sidecarPath` (examples at `README.md:68`, `README.md:103`, `README.md:169`; install docs do the same).
   - `scripts/auto-install.sh` still writes `sidecarPath`.
   - But `src/types.ts` removed `sidecarPath`, and `src/plugin-runtime.ts:63-70` constructs `LibravDBClient` with only `endpoint: cfg.grpcEndpoint`.
   - Result: an existing valid config like:
     ```json
     { "sidecarPath": "unix:/home/computment/.clawdb/run/libravdb.sock" }
     ```
     is accepted by the manifest but ignored by runtime, so the client falls back to `unix:$HOME/.libravdbd/run/libravdb.sock`.
   - I verified this against our isolated Libra profile: it currently has only `sidecarPath: unix:/home/computment/.clawdb/run/libravdb.sock`; the PR client default resolves to `unix:/home/computment/.libravdbd/run/libravdb.sock`, which is not the configured socket here.
   - Fix: either preserve `sidecarPath` as a compatibility alias for `grpcEndpoint`, or remove/migrate it consistently across manifest, docs, installer, config examples, and release notes. Silent ignore is the worst option.

2. Blocking for remote/TLS deployments: `tcp:` endpoints are always converted to `http://`, even when TLS mode resolves to TLS.
   - `src/libravdb-client.ts:102` does:
     ```ts
     const targetUrl = isUnix ? "http://localhost" : rawEndpoint.replace(/^tcp:/, "http://");
     ```
   - `resolveCredentialMode()` can return `"tls"` for remote hosts or forced TLS, and the docs advertise remote TLS/mTLS via `grpcEndpointTlsMode`, `grpcEndpointTlsCa`, and client cert/key.
   - But Connect’s Node gRPC transport selects secure vs plaintext from the `baseUrl` scheme; their own `createGrpcTransport` docs use `https://...` for TLS and `http://...` for plaintext. With this code, `tcp:libravdb.internal:9090` becomes `http://libravdb.internal:9090`, so the TLS options are at best ignored and at worst the client speaks h2c to a TLS daemon.
   - Fix: build `https://host:port` when credential mode is TLS, and `http://host:port` only for insecure/loopback/plaintext. Add tests for `grpcEndpointTlsMode: "tls"` on loopback and `auto` on remote hosts.

3. Blocking for authenticated deployments: `BootstrapSessionKernel` is globally exempted from HMAC signing, and startup auth uses that same stateful method.
   - The interceptor skips auth whenever `req.method.name === "BootstrapSessionKernel"` at `src/libravdb-client.ts:133`, `src/libravdb-client.ts:143`, and `src/libravdb-client.ts:155`.
   - `bootstrapHandshake()` then calls `bootstrapSessionKernel({ sessionId: "auth-bootstrap", userId: "system" })` at `src/libravdb-client.ts:180-190`.
   - The real context-engine bootstrap path also calls `client.bootstrapSessionKernel(...)`, so with `LIBRAVDB_AUTH_SECRET` configured those real session bootstrap requests will also be unsigned forever.
   - The PR description says the HMAC auth handshake runs against `/Health`, and the client even exposes `health()`, but startup never calls it.
   - Fix: use a side-effect-free unauthenticated challenge call (probably `Health`, if that is the daemon contract), then sign every stateful RPC including real `BootstrapSessionKernel`. Add an auth interceptor test that proves normal bootstrap requests carry HMAC after the initial challenge.

Gate results:
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm install --frozen-lockfile` passes.
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm run check` passes: plugin inspector, typecheck, test compile, 105 unit tests.
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm run build` passes.
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm run test:integration` fails one markdown-ingest test: `resume cursor invalidates when target file is deleted during pause`. I reran that exact test on `upstream/main` and it fails there too, same assertion, so I am treating it as inherited broken coverage rather than a regression in this PR.

I did not install this PR into the isolated Libra profile or restart the Libra gateway/daemon. The source review already shows the PR would ignore that profile's current `sidecarPath` socket config, and restarting into a known-bad config would just prove the obvious while making the beta profile noisier.

– Vale

PR #239, opened 2026-05-22
Fix oversized after-turn payloads tripping LibraVDB embedder

fuller-stack-dev:
## Summary
- Bound after-turn payloads before sending them to the daemon ingest path.
- Preserve the original message size for predictive compaction decisions.
- Add a regression test for oversized after-turn content.

## Why
On a local install with `@xdarkicex/openclaw-memory-libravdb@1.6.8` and `libravdbd 1.4.73`, existing memory search still worked, but fresh embedding/reindex hit `context deadline exceeded` followed by `embedding engine not ready`. The failing records were large after-turn payloads. Keeping daemon ingest bounded prevents future oversized turns from poisoning local ONNX embedding while still triggering compaction from the original size.

## Tests
- `./node_modules/.bin/tsc --noEmit`
- `npm run build`
- `./node_modules/.bin/tsc -p tsconfig.tests.json && node --test --test-name-pattern "bounds oversized afterTurn" .ts-build/test/unit/context-engine.test.js`
- `./node_modules/.bin/tsc -p tsconfig.tests.json && node --test .ts-build/test/unit/context-engine.test.js`
- `./node_modules/.bin/tsc -p tsconfig.tests.json && node --test .ts-build/test/integration/host-flow.test.js`

Local note: `node --test .ts-build/test/unit/*.test.js` has one environment-sensitive failure on this machine because a live local daemon makes `bootstrapHandshake wraps transport errors` succeed instead of reject.

## Summary by CodeRabbit

* **Bug Fixes**
  * Messages exceeding size limits are now automatically truncated in after-turn flows with warning notifications when truncation occurs.

* **Tests**
  * Added test coverage for message truncation behavior in after-turn flows.

[Review Change Stack](https://app.coderabbit.ai/change-stack/xDarkicex/openclaw-memory-libravdb/pull/239?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/context-engine.ts
+++ b/src/context-engine.ts
@@ -43,6 +43,7 @@ const QUOTED_PHRASE_RE = /"([^"]{4,})"|'([^']{4,})'/g;
 const EXACT_RECALL_SEARCH_K = 32;
 const EXACT_RECALL_MAX_TOKENS = 4;
 const RESERVED_CURRENT_TURN_TOKENS = 150;
+const AFTER_TURN_INGEST_MAX_TOKENS = 2048;
 const COMMON_QUERY_WORDS = new Set([
   "what", "does", "mean", "remember", "recall", "about", "this", "that",
   "the", "and", "for", "with", "from", "your", "have", "been", "were",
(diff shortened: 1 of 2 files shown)


compoodment (approved):
No blocking issue found.

The change is narrow: it bounds only the daemon after-turn ingest payload while keeping predictive compaction sizing based on the original unbounded messages. That matches the failure mode described here.

Non-blocking nit: the new regression test uses `"x".repeat(50_000)`, so `oversized.endsWith(messages[0].content)` does not strongly prove tail preservation because every substring is identical. A distinct prefix/body/suffix string would make that assertion sharper.

– Vale

xDarkicex:
Superseded by #250 (clean implementation of afterTurn ingest bounding without reverts).

PR #241, opened 2026-05-22
fix: clear predictive context cache on session lifecycle

### LINKED ISSUE #211: bug: predictiveContextCache is not cleared on session reset/end

xDarkicex:
## Summary

Predictive context cache entries were only deleted when `assemble` consumed them. Sessions ending without a final assemble leaked entries indefinitely, causing memory accumulation and stale data risk on recycled `sessionId` values.

Fixes #211.

## What changed

Three guards against cache leakage:

1. **`bootstrap()`** — deletes any stale cache entry for the sessionId being bootstrapped (prevents recycled sessionIds from picking up old predictions)
2. **Max-size eviction** — if the cache reaches 100 entries, the oldest entry is dropped before inserting a new one (fallback for hosts that don't deliver lifecycle hints)
3. **`dispose()`** — clears the entire cache on engine shutdown (prevents leaks across engine replacements)

## Why this is safe

- Bootstrap deletion is a no-op if the sessionId isn't in the cache
- Max-size guard only fires after 100 concurrent sessions — well above typical usage
- `dispose()` is a ContextEngine interface method, called by the host on plugin shutdown
- All existing tests pass

### DIFF
--- a/src/context-engine.ts
+++ b/src/context-engine.ts
@@ -755,6 +755,7 @@ export function buildContextEngineFactory(
   logger: LoggerLike = console,
 ) {
   const predictiveContextCache = new Map<string, import("./types.js").PredictedContext[]>();
+  const PREDICTIVE_CACHE_MAX_SIZE = 100;
   let cachedIdentity: ResolvedIdentity | null = null;
   let cachedSessionKey: string | undefined;
 
@@ -1040,6 +1041,7 @@ export function buildContextEngineFactory(
     ownsCompaction: true,
     async bootstrap(args: { sessionId: string; sessionKey?: string; userId?: string }) {
       const sessionId = requireSessionId(args.sessionId, "bootstrap");
+      predictiveContextCache.delete(sessionId);
       const userId = resolveUserId({
         userIdOverride: args.userId,
         sessionKey: args.sessionKey,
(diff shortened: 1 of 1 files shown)


compoodment (changes requested):
This is directionally useful, but it does not close #211 as written.

#211 is specifically about clearing `predictiveContextCache` on `before_reset` and `session_end`. This PR clears on the next `bootstrap()`, adds a max-size fallback, and clears on `dispose()`, but the existing lifecycle hooks still only forward hints to the daemon via `runtime.emitLifecycleHint`. They do not touch the in-process context-engine cache.

That leaves the original gap: a reset/end without a subsequent bootstrap can still leave stale predictions in the local map, and a later same-session assemble can consume them. The fix needs a local lifecycle clearing path, plus tests proving ended/reset sessions do not reuse cached predictions.

– Vale

PR #268, opened 2026-05-26
fix: run integration suite in default check

compoodment:
## Summary

Runs the integration suite from the default `check` gate and fixes the drift that exposed.

## Scope

- Add a `clean:test` script and run it before TypeScript test compilation so stale `.ts-build` output cannot execute deleted or renamed tests.
- Make `pnpm run check` run both unit and integration test suites.
- Restore the missing structural `FsDirentLike` type used by `markdown-ingest.test.ts`.
- Update `host-flow.test.ts` expectations for the current replay-safe behavior: daemon-only recall is injected as untrusted prompt context, not replayed as transcript messages.
- Normalize compact result details without serializing absent fields as `undefined`.

## Audit Notes

- This is mostly test-gate hygiene, but it includes one small runtime cleanup in compact result normalization.
- The integration failures were real: `pnpm run test:integration` failed before this patch while `pnpm run check` stayed green because integration was outside the default gate.
- The assertion updates preserve the newer prompt-injection boundary instead of weakening it back into transcript replay.

## Verification

```text
COREPACK_ENABLE_AUTO_PIN=0 pnpm run check
```

Passes locally.

– Vale

## Release Notes

* **Tests**
  * Enhanced integration tests to validate system prompt configuration using pattern matching for more robust verification.
  * Improved test reliability and assertion accuracy across host-flow and ingest test suites.

* **Chores**
  * Refactored build and test script workflow to include TypeScript type checking, unit tests, and integration tests in the `check` command.
  * Added test build artifact cleanup to prevent stale files from affecting test execution.


compoodment (commented):
## Review

Reviewed and pushed a cleanup commit on top.

**Original PR — solid, no bugs found:**

- `clean:test` nuking stale `.ts-build` before compilation is a real fix
- Expanding `check` to include integration is the core claim, works
- `normalizeCompactResult` change correctly eliminates `undefined` keys from `details` — the old code set absent fields to `undefined` explicitly; the new code only assigns when the value exists. Strictly better for `Object.keys()` / `in` checks downstream
- Test assertion updates match current runtime behavior (daemon recall → `memory_item` in `systemPromptAddition`, not transcript replay in `messages`)

**Cleanup commit adds:**

1. Replaced `node -e "require('node:fs').rmSync(...)"` inline CJS require in an ESM package with `scripts/clean-test-build.mjs` — proper ESM, not an inline hack
2. Exported `FsDirentLike` from `src/markdown-ingest.ts` and imported it in the test instead of duplicating the interface

**Minor notes (not blocking):**

- `check` runs `plugin:ci` twice (via `test:ts` + `test:integration`). Could be optimized with a shared setup step, but the complexity isn't worth it for CI
- The `^<recalled_memories>` regex anchor in `host-flow.test.ts` tests the mock's format, not the plugin code. Harmless but slightly redundant alongside the `<memory_item>` assertions

All 166 unit + 46 integration tests pass on `pnpm run check`.

PR #270, opened 2026-05-27
Harden OpenClaw metadata and replay context

### LINKED ISSUE #269: Context engine stalls and replay contamination from OpenClaw metadata
## Summary

The OpenClaw context-engine integration can spend many seconds in LibraVDB pre-provider work when an inbound channel message includes OpenClaw's leading untrusted metadata envelope.

The original symptom looked like a model/provider delay, but instrumentation showed the provider request had not started yet. The delay was in the LibraVDB context engine before prompt submission.

While fixing that, live testing surfaced a related replay-contamination issue: if historical/session-projected context replays tool syntax, tool results, OpenClaw reply directives, empty memory shells, or assistant "let me look that up" preambles, the model can copy those artifacts on later turns instead of invoking tools structurally.

## Affected prompt shape

The affected prompts begin with channel metadata blocks like:

```text
Conversation info (untrusted metadata):
~~~json
{ "chat_id": "channel:...", "message_id": "...", "sender_id": "..." }
~~~

Sender (untrusted metadata):
~~~json
{ "label": "...", "id": "..." }
~~~

Actual user request...
```

OpenClaw may also include reply target, forwarded context, thread starter, and chat-history blocks. These blocks are useful current-turn context, but they are volatile transport metadata and should not be fed to durable-memory ingestion/search as full content.

## Sanitized evidence

No raw trajectory, packet capture, hostnames, local paths, IP addresses, message IDs, sender IDs, emails, usernames, or private channel content are included here.

Observed timing from local instrumentation, redacted and rounded:

```text
context-engine afterTurn:  about 8.3s
context-engine assemble:   about 6.6s
total pre-provider gap:    about 14.9s
session lock/release path: under 15ms
```

A direct sidecar probe showed the same pattern:

```text
short prompt + short user text:       about 0.3-0.6s
metadata envelope + short user text:  about 5-8s
```

After stripping the OpenClaw metadata envelope before sending content to LibraVDB:

```text
afterTurn with same prompt shape: about 0.5s
assemble with same prompt shape:  about 0.4s
```

Additional replay-contamination evidence from sanitized live runs:

```text
- structured tool-call history was transformed into assistant-visible text;
- raw tool-result JSON could be replayed as assistant content;
- OpenClaw reply directives such as [[reply_to_current]] could become provider-visible historical text;
- empty recalled-memory shells appeared even when no useful memory text was present;
- assistant-only lookup preambles primed later turns to say "let me look that up" without making a structural tool call.
```

## Why this matters

This is both a performance issue and a context hygiene issue.

The full OpenClaw metadata envelope is too expensive/noisy for durable memory, but stripping it completely loses useful low-cost routing facts. The safer boundary is:

- ingest/search against the actual user text plus compact routing facts;
- preserve useful channel/chat/sender identifiers needed for continuity;
- avoid replaying private reply/history bodies;
- never replay executable-looking tool syntax or raw tool results as assistant text;
- keep recalled memory in clearly untrusted system context rather than as fake transcript turns.

The issue is channel-agnostic. It can affect Discord, Telegram, iMessage, and any OpenClaw channel that prepends the same inbound metadata blocks.

## Current fix in PR #270

PR #270 now:

- strips full OpenClaw metadata envelopes before `afterTurn` ingestion and assemble-time lookup;
- retains compact Discord/iMessage routing context without retaining private quote/history bodies;
- sanitizes OpenClaw reply directives from historical assistant replay;
- removes structured tool calls, raw tool-result JSON, executable-looking tool syntax, and assistant lookup preambles from replay;
- removes empty recalled-memory shells;
- adds a generic recall fallback when daemon projection is empty but durable search has usable result text;
- hydrates recall candidates from metadata text when search result `text` is empty;
- adds focused regression tests around all of the above.

Latest verification from the PR branch:

```text
pnpm run test:ts: 187 tests passed
pnpm run build: passed
git diff --check: clean
```

JuanHuaXu:
## Summary

Fixes #269.

This PR started as a performance fix for OpenClaw inbound metadata envelopes, but the first fix surfaced a deeper replay-contamination problem in the LibraVDB context-engine boundary.

It now hardens the OpenClaw context-engine integration across three related surfaces:

- prevent full OpenClaw inbound metadata envelopes from being ingested or used as expensive retrieval prompts;
- preserve a compact, low-cost subset of useful routing context for Discord/iMessage continuity;
- prevent recalled/session-projected context from reintroducing executable-looking tool syntax, tool-result JSON, OpenClaw reply directives, empty recalled-memory shells, or assistant "let me look that up" preambles that prime later turns to stop using tools structurally.

## Root cause

OpenClaw channel messages can begin with AI-facing transport metadata such as conversation info, sender info, reply target, and chat-history blocks. Those blocks are useful for the live turn, but they are volatile transport metadata and are expensive/noisy durable-memory input.

The original envelope behavior caused multi-second pre-provider stalls because LibraVDB was asked to ingest/search over full channel envelopes instead of the actual user request.

After the envelope strip landed locally, live testing exposed a buried second-order issue: historical replay could still surface stale assistant text and raw tool artifacts as provider-visible context. That primed the model to copy old tool syntax, repeat stale tool preambles, emit marker-only replies, or narrate "let me look that up" instead of making structural tool calls.

## Changes

- Strip leading OpenClaw untrusted metadata blocks before `afterTurn` ingestion.
- Strip/normalize the same envelope before `assemble` prompt lookup and exact-recall fallback.
- Retain only compact OpenClaw context fields useful for continuity, such as provider/channel labels, channel/chat ids, server id, sender id/name, and iMessage chat identifiers.
- Drop reply target and chat-history bodies from retained context so private quoted/history text is not replayed as memory content.
- Sanitize OpenClaw reply directives such as `[[reply_to_current]]`, `[[reply_to:<id>]]`, and `[[audio_as_voice]]` from historical assistant replay while preserving useful answer text.
- Omit historical structured tool-call blocks and raw tool-result messages from assistant replay.
- Strip executable-looking tool-call JSON/bracket syntax from recalled memory and assemble results.
- Drop assistant-only lookup preambles, including multi-sentence forms such as "Let me look that up for you. Let me look up the current price."
- Remove empty `<recalled_memories>` shells so blank recall projections do not look like useful memory context.
- Add a generic recall fallback for memory-style queries when daemon projection is empty but durable search has usable text in result metadata.
- Hydrate recall candidates from metadata text when the top-level search result text is empty.
- Add focused regression tests for Discord/iMessage compact context retention, tool/directive/preamble sanitization, empty recall cleanup, and metadata-backed generic recall.

## Sanitized evidence

Raw packet captures, trajectory exports, and local session files are intentionally not attached because they can contain prompts, tool schemas, local paths, message IDs, sender identifiers, and private channel content.

Redacted timing from local instrumentation before the first fix:

```text
context-engine afterTurn:  about 8.3s
context-engine assemble:   about 6.6s
total pre-provider gap:    about 14.9s
session lock/release path: under 15ms
```

Direct sidecar probe of content shape:

```text
short prompt + short user text:       about 0.3-0.6s
metadata envelope + short user text:  about 5-8s
```

After stripping full envelopes before LibraVDB RPCs:

```text
afterTurn with same prompt shape: about 0.5s
assemble with same prompt shape:  about 0.4s
```

Subsequent sanitized live observations that motivated the expanded scope:

```text
- prior assistant tool syntax was visible in later provider context before a tool call loop;
- prior assistant reply directives were visible before a marker-only final answer;
- blank recalled-memory shells were injected even when no memory item text was present;
- a current-price request replayed old "let me look up" assistant text instead of a structural tool call.
```

## Verification

Latest local verification on this branch:

```bash
pnpm run test:ts
pnpm run build
git diff --check
```

Observed:

```text
pnpm run test:ts: 187 tests passed
pnpm run build: passed
git diff --check: clean
```

Latest pushed commit: `1865de2 fix: harden context replay recall`.

## Privacy note

The PR body and issue intentionally use only redacted examples and generic identifiers. The raw repro artifacts are treated as sensitive because OpenClaw trajectory bundles and packet captures can contain prompts, model messages, tool schemas, tool results, runtime paths, local host details, and private channel metadata.


compoodment (changes requested):
I tested this at head `6773c451b34b7904ecdfa79cffaf10bd92ac1d3b`.

The normal-path fix looks good: the OpenClaw envelope is stripped before `afterTurn` ingest, stripped from the `assemble` prompt, and exact-recall lookup no longer searches metadata tokens. A live sidecar timing probe also no longer showed the metadata-envelope penalty after warmup.

Blocking issues I found:

1. `pnpm run test:integration` still fails on this PR head. `pnpm run test:ts` passes (168/168) and `pnpm run check` passes, but `check` does not include integration tests. The integration failures are all in `test/integration/host-flow.test.ts`:
   - assemble mapping returns an extra `<retrieved_memory>` block in `systemPromptAddition`
   - compaction-related assemble paths return `undefined` message content where the tests expect assembled messages
   - compact result normalization includes `summaryText: undefined` in `details`

   I also checked base `main`: it currently fails earlier at TS compile because `markdown-ingest.test.ts` is missing `FsDirentLike`. This PR fixes that compile issue, but once integration actually runs, PR head still fails.

2. The fallback/error paths still use unsanitized `args.messages`. The normal daemon path uses normalized `messages`, but the fallback paths call `buildBudgetFallbackContext(args.messages, ...)` / `ensureReplaySafeUserTurn(..., args.messages, ...)`. I verified this with a fake-client probe by forcing `assembleContextInternal` to throw: the returned fallback message still began with the OpenClaw metadata envelope. That means provider-facing fallback context can still replay the exact metadata this PR is trying to keep out.

Suggested fix: use the already-normalized `messages` in the fallback paths too, and add a regression test where daemon assembly throws while the user message starts with an OpenClaw metadata envelope.

Existing CodeRabbit discussion: the header-only envelope issue appears addressed by the latest commit, so I am not re-raising that one.

JuanHuaXu:
## Maintainer note: second fix documented, and what it does/does not cover

This PR now covers two related context-engine issues that were found in sequence.

1. Metadata-envelope stall/data-hygiene fix

The first fix strips OpenClaw's leading inbound metadata envelopes before durable-memory ingestion, prompt-based context lookup, exact-recall fallback, and fallback replay paths. That removed the observed pre-provider stall and stopped volatile transport metadata from becoming durable memory input.

2. Tool-call replay contamination fix

After the first fix made the prompt path fast enough to inspect clearly, it surfaced a buried replay problem: structured OpenClaw tool activity could be flattened back into assistant text shaped like executable tool syntax, for example:

```text
[tool:web_search] {"query":"...","count":5}
```

That is dangerous because a later model turn can imitate the leaked text instead of invoking tools through the structured tool-call channel.

The latest branch patch addresses this inside the LibraVDB context engine by:

- converting structured `toolCall` / `toolResult` history into inert historical summaries;
- replacing bracketed, JSON-ish, and XML-ish tool-call syntax before provider-visible replay;
- avoiding raw tool-result JSON blobs as assistant transcript messages;
- keeping the latest real user turn intact;
- adding regression coverage for block normalization, tool-result projection, assemble/fallback replay, and provider-visible contamination checks.

What this does not claim to fix:

A later runaway web-search case occurred after compiled context was clean: the model repeatedly made structural `web_search` calls because search results were ambiguous and OpenClaw's loop detector only warned at the configured thresholds. That is a separate search-control / harness-loop issue, not evidence that LibraVDB is still leaking `[tool:...]` text.

Sanitization note: raw trajectories and local captures were not attached because they can include prompts, tool schemas, local paths, message IDs, sender identifiers, and other private runtime details.

compoodment (changes requested):
Re-reviewed and tested head `3aaa0e70898e1257532b610ec395248db786c208`.

Good news: my previous blockers are fixed.

Verification run:
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm run test:ts`: pass, 180/180
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm run test:integration`: pass, 46/46
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm run check`: pass

The metadata fallback leak I called out previously is covered now, and the old integration failures in `host-flow.test.ts` are gone.

Remaining blocker:

The new assistant/tool sanitization path strips ordinary assistant JSON that merely contains a `"name"` field. Because `normalizeKernelMessage` now sanitizes every assistant message before assembly/ingest, this is no longer limited to recalled tool-looking blobs.

Current head behavior I reproduced:

```js
normalizeKernelMessages([
  { role: "assistant", content: 'Return JSON like {"name":"Alice","age":30} exactly.' }
])
```

returns:

```js
[{ role: "assistant", content: "Return JSON like exactly." }]
```

Same issue appears when daemon-recalled assistant content is wrapped into `<retrieved_memory>`: the JSON object is removed from the memory item.

Root cause is that `TOOL_CALL_JSON_RE` matches any JSON object with a `"name"` property, not specifically a tool-call shape. Then `sanitizeToolCallPatterns` removes it from mixed prose. A normal assistant explanation of JSON, schema examples, person records, API payloads, etc. can be silently corrupted before durable-memory ingestion or replay.

Suggested fix: narrow JSON tool-call detection to actual tool-call shapes, e.g. require an `arguments` / `parameters` field or another OpenAI/OpenClaw tool-call-specific marker, and add a regression test that preserves ordinary JSON like `{"name":"Alice","age":30}` while still removing executable tool-call JSON.

Non-blocking hygiene: there is still a `console.log("DEBUG_SP_OUTPUT", ...)` in `test/unit/context-engine.test.ts`, which is why the passing test run prints debug JSON. CodeRabbit already flagged that, so I am not treating it as the substantive blocker.

compoodment (dismissed):
Re-reviewed latest head `8ad9fb2`.

No blocking findings. The latest changes cover the prior review points I cared about:
- leading OpenClaw metadata envelopes are stripped before afterTurn and assemble
- header-only metadata blocks are covered
- tool call/result replay is dropped from normalized transcript paths
- assistant prose around embedded tool-looking syntax is preserved
- recalled memory/tool-ish content is moved into guarded system prompt context instead of inline replay
- stale .ts-build cleanup is wired into the test scripts

Verified locally in a temporary worktree:
- COREPACK_ENABLE_AUTO_PIN=0 pnpm install --frozen-lockfile --ignore-scripts
- COREPACK_ENABLE_AUTO_PIN=0 pnpm run test:ts — 183 passed
- COREPACK_ENABLE_AUTO_PIN=0 pnpm run test:integration — 46 passed
- COREPACK_ENABLE_AUTO_PIN=0 pnpm run build — passed

Non-blocking note: the tool sanitizer now drops tool activity entirely via formatHistoricalToolActivity() => "", despite some comments still saying "neutral summary". That is fine for the replay-safety goal. In an extra probe, deeply nested tool-call JSON can leave a stray punctuation fragment like `,"count":5}}`, but the executable tool name/query payload is gone and I do not see a practical replay path there.

JuanHuaXu:
## Follow-up: preserving live tool results after replay sanitization

The replay-safety work in this PR intentionally prevents historical tool calls/results from being reintroduced as provider-visible assistant text. Follow-up testing exposed a separate edge in that fix: the sanitizer was also too aggressive for the live current-turn tool protocol tail.

What happened:

- Historical tool activity must be sanitized or omitted so it cannot prime the model with executable-looking tool syntax.
- Live current-turn tool calls/results are different: after a model issues a structural tool call, the next provider pass needs the corresponding live tool result so it can synthesize the final answer.
- The replay sanitizer could remove that live tool result along with historical replay content, which made the model lose the result it had just requested.

Latest fixes included here:

- `64cd501` restores only the live current-turn tool protocol tail after context assembly, so post-tool synthesis still receives the fresh tool result while historical replay remains sanitized.
- `082ac00` adds a final token-budget clamp after restoring that live tail, so a large restored tool result cannot bypass the existing budget invariant.

Validation after these fixes:

```text
pnpm run test:ts          184 passed
pnpm run test:integration 46 passed
pnpm run build            passed
git diff --check          passed
```

This is a narrow correction to the replay-safety patch: historical tool contamination remains blocked, live tool-result continuity is preserved, and restored live tool content remains subject to the final context budget.

PR #271, opened 2026-05-28
Register LibraVDB memory tools

fuller-stack-dev:
## Summary
- register LibraVDB-owned `memory_search` and `memory_get` tools backed by the existing runtime bridge
- declare the tool contracts in `openclaw.plugin.json` so installation/tool discovery can see them without enabling `memory-core`
- add coverage for search/get registration, session corpus filtering, and safe rejection of unreturned paths
- add the missing markdown-ingest test-only `FsDirentLike` type so the full test TypeScript compile passes

## Verification
- `npm run build`
- `./node_modules/.bin/tsc -p tsconfig.tests.json --pretty false`
- `node --test .ts-build/test/unit/*.test.js`
- `node --test .ts-build/test/integration/checklist-validation.test.js .ts-build/test/integration/markdown-ingest.test.js`
- `node_modules/.bin/plugin-inspector ci --no-openclaw --runtime --mock-sdk --allow-execute`

## Notes
- Plugin Inspector passes with its existing P2 inspector-gap suggestion: isolated cold import requires dependency installation.
- The broader integration command including `host-flow.test.js` still has pre-existing host-flow expectation failures unrelated to this tool registration path.

## Summary by CodeRabbit

* **New Features**
  * Memory tools (memory_search, memory_get) are now available and routed through the LibraVDB-backed daemon for direct use.
  * Memory-related guidance now prompts using memory_search first, then memory_get.

* **Documentation**
  * Highlights updated to reflect the new memory tools integration.

* **Reliability**
  * Search results and excerpt reads now fall back to metadata-derived text when backend text is missing, improving result accuracy.

[Review Change Stack](https://app.coderabbit.ai/change-stack/xDarkicex/openclaw-memory-libravdb/pull/271?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/openclaw.plugin.json
+++ b/openclaw.plugin.json
@@ -9,7 +9,11 @@
   ],
   "contracts": {
     "memory": true,
-    "contextEngine": true
+    "contextEngine": true,
+    "tools": [
+      "memory_search",
+      "memory_get"
+    ]
   },
   "activation": {
     "onCommands": [
--- a/src/index.ts
+++ b/src/index.ts
@@ -69,7 +69,8 @@ export function register(api: OpenClawPluginApi) {
     : createPluginRuntime(cfg, logger);
   registerMemoryCli(api, runtimeOrNull, cfg, logger);
 
-  if (runtimeOrNull) {
+  const ownsMemorySlot = memSlot === MEMORY_ID;
+  if (runtimeOrNull && ownsMemorySlot) {
     const memoryTools = createLibraVdbMemoryTools(runtimeOrNull.getClient, cfg, logger);
     api.registerTool?.((ctx) => memoryTools.createSearchTool(ctx), { names: ["memory_search"] });
     api.registerTool?.((ctx) => memoryTools.createGetTool(ctx), { names: ["memory_get"] });
--- a/src/index.ts
+++ b/src/index.ts
@@ -7,6 +7,7 @@ import { createDreamPromotionHandle } from "./dream-promotion.js";
 import { createMarkdownIngestionHandle } from "./markdown-ingest.js";
 import { buildMemoryPromptSection } from "./memory-provider.js";
 import { buildMemoryRuntimeBridge } from "./memory-runtime.js";
+import { createLibraVdbMemoryTools } from "./memory-tools.js";
 import { createPluginRuntime } from "./plugin-runtime.js";
 import type { PluginConfig } from "./types.js";
 
@@ -67,6 +68,12 @@ export function register(api: OpenClawPluginApi) {
     ? null
     : createPluginRuntime(cfg, logger);
   registerMemoryCli(api, runtimeOrNull, cfg, logger);
+
+  if (runtimeOrNull) {
+    const memoryTools = createLibraVdbMemoryTools(runtimeOrNull.getClient, cfg, logger);
+    api.registerTool?.((ctx) => memoryTools.createSearchTool(ctx), { names: ["memory_search"] });
+    api.registerTool?.((ctx) => memoryTools.createGetTool(ctx), { names: ["memory_get"] });
+  }
 
   if (isLightweight || isDiscovery) {
     if (!isLightweight) {
--- a/src/index.ts
+++ b/src/index.ts
@@ -14,7 +14,7 @@ import type { PluginConfig } from "./types.js";
 export const MEMORY_ID = "libravdb-memory";
 
 const LIGHTWEIGHT_MODES = new Set(["cli-metadata", "setup-only"]);
-const RUNTIME_CLEANUP_SHUTDOWN_REASONS = new Set(["delete", "restart"]);
+const RUNTIME_CLEANUP_SHUTDOWN_REASONS = new Set(["delete"]);
 
 export function shouldShutdownRuntimeForLifecycleCleanup(reason: string): boolean {
   return RUNTIME_CLEANUP_SHUTDOWN_REASONS.has(reason);
--- a/src/memory-provider.ts
+++ b/src/memory-provider.ts
@@ -16,6 +16,8 @@ function buildToolGuidance(availableTools: ReadonlySet<string> | undefined): str
   const lines = [
     "For explicit memory lookup requests, call `memory_search` first.",
     "Use it for prior turns, remembered facts, earliest interactions, and channel history.",
+    "Do not answer memory lookup requests from prior transcript claims or earlier `memory_search` results; perform a fresh `memory_search` for the current request.",
+    "For earliest or oldest memory questions, request enough results, compare timestamps in the returned snippets, and use `memory_get` if the snippet is not enough.",
   ];
   if (availableTools.has("memory_get")) {
     lines.push("After a `memory_search` hit, call `memory_get` when exact wording or more context is needed.");
--- a/src/memory-provider.ts
+++ b/src/memory-provider.ts
@@ -9,16 +9,34 @@ const MEMORY_PROMPT_HEADER = [
   "",
 ] as const;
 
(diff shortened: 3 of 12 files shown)


compoodment (changes requested):
Reviewed latest head `194a21b`.

Requesting changes for one correctness issue.

Blocking:
- `src/memory-tools.ts:168-174`: `corpus` filtering is applied after `manager.search()` already queried the mixed collection set. For `corpus: "sessions"`, `buildSearchContext(ctx)` still gives the runtime bridge enough context to search session + user + global with the requested `k`; only then `filterResultsByCorpus` drops non-session rows. That means durable-memory rows can fill the top-k and cause the tool to return no session results even when matching session rows exist below the cut.

Focused repro:
- `memory_search({ query: "needle", corpus: "sessions", maxResults: 1 })`
- RPC call observed: `searchTextCollections({ collections: ["session:s1", "user:u1", "global"], k: 1 })`
- when the top returned row was `collection: "user:u1"`, the tool response was `{ results: [] }`

This violates the advertised corpus semantics. `sessions` and `memory` need to constrain the searched collections before ranking/top-k, not filter after ranking. Fix options: pass a corpus/source selector through the runtime bridge, or have the tool choose the collection set for `sessions`/`memory` before the search. Add a regression where durable memory outranks a session hit but `corpus: "sessions"` still returns the session hit.

Notes:
- The latest `194a21b` slot-gating update looks directionally right.
- I did not treat CodeRabbit's rejected manager-promise cache comment as blocking; `getMemorySearchManager()` currently wraps status/getClient failures into a manager with disabled status, so I do not see a practical poisoned-promise path in current code.

Verification:
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm install --frozen-lockfile --ignore-scripts`
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm run test:ts` - pass, 169/169
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm run build` - pass
- focused `node --test .ts-build/test/integration/checklist-validation.test.js .ts-build/test/integration/markdown-ingest.test.js` - pass, 26/26
- full `COREPACK_ENABLE_AUTO_PIN=0 pnpm run test:integration` - fails 42/46, in existing host-flow expectations. That still needs a clean rebase/stacking story, but it is separate from the corpus bug above.

compoodment (changes requested):
Re-reviewed through `4105fd7` (latest head).

**Three new commits since my last review** (`194a21b`):

- `b48f748` — Tool guidance in `memory-provider.ts`: steers recall through `memory_search`/`memory_get`. Fine addition, tests updated accordingly.
- `68aa625` — `memory-runtime.ts`: hydrates blank `text` from `metadataJson.text` via `resolveSearchResultText`. Good fix — the runtime bridge and tool path both benefit. Also changes `RUNTIME_CLEANUP_SHUTDOWN_REASONS` from `["delete", "restart"]` to `["delete"]` — this means a `restart` lifecycle event no longer shuts down the runtime. Intentional? If so, worth a one-liner in the commit message; it is a behavior change that could affect runtime teardown during hot restarts.
- `4105fd7` — Adds "perform a fresh `memory_search`" guidance line to `memory-provider.ts`, minor `memory-tools.ts` param description tweak. Clean.

**The blocking corpus bug is still unfixed.**

`src/memory-tools.ts` lines ~158-175: `buildSearchContext(ctx)` passes the full context (agentId, sessionId, sessionKey) into `manager.search()`, which queries all collections (`session:*`, `user:*`, `global`). Then `filterResultsByCorpus` strips results **after** top-k has already been applied.

Concrete repro:
```ts
// memory_search({ query: "needle", corpus: "sessions", maxResults: 1 })
// RPC: searchTextCollections({ collections: ["session:s1", "user:u1", "global"], k: 1 })
// → top hit is collection "user:u1" with score 0.92
// → filterResultsByCorpus drops it (source !== "sessions")
// → returns { results: [] } even though a matching session row exists below the top-k cutoff
```

This violates the advertised corpus semantics. `corpus: "sessions"` and `corpus: "memory"` need to constrain the searched collections **before** ranking, not filter after.

**Fix options:**
1. Pass a corpus-aware collection selector through the runtime bridge so `searchTextCollections` only receives the relevant collection set.
2. Have the tool compute the collection set from the corpus flag before calling `manager.search()`, and pass it as a parameter.

Option 1 is cleaner because it keeps the runtime bridge's existing `buildSearchContext` contract and moves the collection decision to where the RPC call is actually shaped. Either way, add a regression test where durable-memory rows outrank a session hit but `corpus: "sessions"` still returns the session hit.

**Non-blocking notes:**

- CodeRabbit's `.catch` eviction suggestion on `getManager`: I already noted in my first review that `getMemorySearchManager()` wraps failures into a manager with disabled status rather than throwing, so a poisoned-promise path does not exist in current code. Still a reasonable defensive measure but not blocking.
- `RUNTIME_CLEANUP_SHUTDOWN_REASONS` dropping `"restart"`: call this out explicitly if intentional.
- Test file `markdown-ingest.test.ts` adds a local `FsDirentLike` type. Fine for test-only, but the PR description says it is fixing a compile issue — worth confirming this is not papering over a missing export that other consumers might also hit.

Requesting changes for the corpus bug. Everything else looks solid.

fuller-stack-dev:
Addressed the review comments in 92fc23f.

- `corpus` now flows from `memory_search` into the runtime bridge and constrains the daemon query before top-k/ranking. `sessions` searches only the resolved session collection, and `memory` searches only durable `user:`/`global` collections.
- Rejected memory manager initialization promises are evicted from the cache so a later call can retry instead of reusing a failed promise.
- Added regressions for session and durable-memory corpus scoping before ranking.

Verification:
- `npm run build`
- `./node_modules/.bin/tsc -p tsconfig.tests.json --pretty false`
- `node --test .ts-build/test/unit/*.test.js`
- `node --test .ts-build/test/integration/checklist-validation.test.js .ts-build/test/integration/markdown-ingest.test.js`
- `npm run plugin:ci`

compoodment (dismissed):
Re-reviewed through `92fc23f` (latest head).

**Corpus bug is fixed.** `resolveSearchCollections` now scopes the collection set *before* the RPC call based on the `corpus` parameter:
- `corpus: "sessions"` → only `[resolveSessionSearchCollection(...)]` (or `[]` when no session context)
- `corpus: "memory"` → only `[resolveUserCollection(...), "global"]`
- `corpus: "all"` (default) → full set including session + durable

Dream queries are correctly skipped when `corpus === "sessions"` (dream queries search durable memory, not sessions).

The regression tests are well-structured:
- `CorpusPriorityRpc` returns a durable row at 0.99 score, asserting the sessions corpus path calls `searchText` on a single session collection — not `searchTextCollections` with mixed collections. This is exactly the repro scenario from the bug.
- The `memory` corpus test asserts `searchTextCollections` receives only `["user:u1", "global"]`.

`filterResultsByCorpus` remains as a post-filter safety net, which is fine — now that collections are pre-scoped, the post-filter is redundant but harmless defensive code.

Also verified:
- `getManager` now has the `.catch` eviction for rejected promises. Good defensive addition.
- `RUNTIME_CLEANUP_SHUTDOWN_REASONS` dropping `"restart"` is self-documented by the commit message ("preserve runtime").
- `FsDirentLike` in `markdown-ingest.test.ts` is test-only, fine.

**All gates pass:**
- `pnpm run build` ✅
- `pnpm run test:ts` — 172/172 ✅
- `pnpm run check` — 131/131 ✅
- Integration: 26/26 ✅
- Memory-tools: 5/5 ✅ (including both corpus-priority regressions)

Approved.

PR #287, opened 2026-06-01
fix: cap compaction threshold at compactSessionTokenBudget to prevent Codex Runtime blowout

xDarkicex:
## Bug

Codex Runtime passes a large `tokenBudget` (100k+) to `assemble()`. The dynamic compaction threshold becomes `0.8 * 100k = 80k tokens` — far above the configured `compactSessionTokenBudget` (default 2000). The session never triggers auto-compaction because it can't reach the inflated threshold.

## Fix

Cap the derived threshold at `compactSessionTokenBudget` in `resolveDynamicCompactThreshold`. Explicit `compactThreshold` config still wins when set.

## Test plan
- [ ] Session over 2000 tokens with `compactSessionTokenBudget=2000` triggers auto-compact
- [ ] Codex Runtime with `tokenBudget=100000` still caps threshold at 2000
- [ ] Explicit `compactThreshold=5000` overrides the cap

## Summary by CodeRabbit

* **Chores**
  * Updated dependency version for contracts library

* **Refactor**
  * Enhanced internal tracing and logging capabilities for compaction operations, including daemon state visibility and dynamic threshold configuration with budget capping

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,6 +79,6 @@
   "dependencies": {
     "@connectrpc/connect": "^1.7.0",
     "@connectrpc/connect-node": "^1.7.0",
-    "@xdarkicex/libravdb-contracts": "^2.0.19"
+    "@xdarkicex/libravdb-contracts": "^2.0.20"
   }
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -15,8 +15,8 @@ importers:
         specifier: ^1.7.0
         version: 1.7.0(@bufbuild/protobuf@1.7.2)(@connectrpc/connect@1.7.0(@bufbuild/protobuf@1.7.2))
       '@xdarkicex/libravdb-contracts':
-        specifier: ^2.0.19
-        version: 2.0.19
+        specifier: ^2.0.20
+        version: 2.0.20
     devDependencies:
       '@bufbuild/protobuf':
         specifier: 1.7.2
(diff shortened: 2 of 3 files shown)


xDarkicex:
Clean, focused PR. The fix is correct and the logging is backward-compatible.

**Core fix — `resolveDynamicCompactThreshold`:**

The pre-fix code returned `floor(budget * fraction)` with no bounds. A 100k budget × 0.8 = 80k threshold — unreachable for most sessions. The `[2000, 16000]` clamp plus `compactSessionTokenBudget` cap fixes this. Explicit `compactThreshold` still wins (branch 1), which is the right precedence.

**Daemon state logging — `normalizeCompactResult`:**

All four new fields are guarded with proper type checks (`bigint` for int64, `number` for int32, `boolean` for bool). The spread into `details` converts `bigint` to string to avoid JSON serialization issues. Null-safe — old daemons that don't populate these fields produce no log and no detail entries.

**One observation, not blocking:**

The trace logging in `resolveDynamicCompactThreshold` requires a `logger` parameter, but `getDynamicCompactThreshold` (the arrow wrapper) never passes one. So the `[compact:trace]` lines are currently dormant — they only fire if someone calls `resolveDynamicCompactThreshold` directly with a logger. Consider wiring the logger through `getDynamicCompactThreshold` in a follow-up so all compaction code paths produce threshold traces.

LGTM.

PR #301, opened 2026-06-02
feat: expose graph edge metadata (caused_by, leads_to, timestamp) in memory_search

xDarkicex:
## Summary
- Surface causal graph edge metadata from `metadataJson` in `memory_search` results
- `why_ids` → `caused_by` (upstream causal parents — "from")
- `how_ids` → `leads_to` (downstream procedural children — "to")
- `ts` → `timestamp` (ISO 8601, human-readable)
- Also exposes `memory_kind` and `memory_signals` in structured result path (was only in legacy path)

No daemon changes required — data already flows through `metadataJson`, just wasn't surfaced in tool output.

## Test plan
- [ ] Verify `memory_search` returns `caused_by`/`leads_to` arrays on records with graph edges
- [ ] Verify `memory_search` returns `timestamp` on records with `ts` metadata
- [ ] Verify `memory_kind` and `memory_signals` appear in structured results
- [ ] Verify legacy path returns same fields

## Summary by CodeRabbit

* **New Features**
  * Memory search results now show enriched metadata: kind, signals, causal links (what caused / what leads to), and normalized timestamps for clearer context.
* **Documentation**
  * Help/description for the memory search tool updated to document the new fields and clarify usage guidance.

### DIFF
--- a/src/memory-runtime.ts
+++ b/src/memory-runtime.ts
@@ -131,11 +131,22 @@ function createMemorySearchManager(
       const legacyResults = filteredResults.map((item) => {
         const meta = parseMetadataJson(item);
         const text = resolveSearchResultText(item, meta);
-        return {
+        const kind = typeof meta.memory_kind === "string" ? meta.memory_kind : undefined;
+        const signals = meta.memory_signals as string[] | undefined;
+        const causedBy = stringArrayFromMeta(meta, "why_ids");
+        const leadsTo = stringArrayFromMeta(meta, "how_ids");
+        const timestamp = metaTimestamp(meta);
+        const enriched: Record<string, unknown> = {
           ...item,
           text,
           content: text,
         };
+        if (kind) enriched.kind = kind;
+        if (signals && signals.length > 0) enriched.signals = signals;
+        if (causedBy && causedBy.length > 0) enriched.caused_by = causedBy;
+        if (leadsTo && leadsTo.length > 0) enriched.leads_to = leadsTo;
+        if (timestamp) enriched.timestamp = timestamp;
+        return enriched;
       });
       if (legacyCall) {
         return { results: legacyResults };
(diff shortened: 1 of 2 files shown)


compoodment:
## Vale Review

Verdict: **not merge-ready**. The runtime enrichment idea is sound, but this branch currently conflicts with `main` and the public tool contract has two sharp edges that will mislead agents.

### Findings

1. **Blocker: this PR conflicts with current `main` and would drop newer `memory_search` guidance.**

GitHub reports the PR as `CONFLICTING`. I also checked with `git merge-tree upstream/main upstream/pr-301`, which produces a content conflict in `src/memory-tools.ts`.

The conflict matters: current `main` has tightened the `memory_search` description with the anti-loop instruction:

`Do NOT call memory_search if the answer is already visible in your context window...`

The PR replaces that description with the new result-field docs. When rebasing, keep both pieces: the anti-loop guidance from `main` and the new metadata-field documentation.

2. **Major: the docs tell callers to pass graph edge IDs into `memory_get`, but `memory_get` accepts only search-result paths.**

In `src/memory-tools.ts:202-203`, the new description says:

- `caused_by`: “Use memory_get with these IDs...”
- `leads_to`: “Use memory_get with these IDs...”

That is wrong for the actual tool contract. `MEMORY_GET_SCHEMA` requires `path`, and its description says “A path returned by memory_search.” Raw `why_ids` / `how_ids` are not encoded search paths and are not inserted into `returnedSearchPaths` by `memory-runtime.ts`, so `memory_get` will reject them with the existing hidden-path guard.

Fix direction: either expose resolvable paths alongside the graph IDs, or describe these IDs as correlation/search keys, not direct `memory_get` inputs.

3. **Medium: `memory_signals` is surfaced with a type assertion instead of runtime shape checking.**

In `src/memory-runtime.ts:135` and `src/memory-runtime.ts:321`:

```ts
const signals = meta.memory_signals as string[] | undefined;
```

Then the code checks `signals.length`. If metadata contains a string, this leaks a string into `signals` even though the documented result field is an array/bitmask concept. If metadata contains a different non-array shape, behavior becomes inconsistent.

Use the same defensive pattern as `why_ids` / `how_ids`, or explicitly support the real stored shape if `memory_signals` is not always `string[]`.

4. **Test gap: the new output contract is not locked down.**

This PR changes both legacy and structured result paths, but adds no tests. I would want focused tests for:

- legacy `manager.search("query")` returning `kind`, `signals`, `caused_by`, `leads_to`, and `timestamp`
- structured `manager.search({ query })` returning the same fields
- malformed metadata does not leak bad shapes into `signals`
- valid numeric `ts` and ISO-ish string timestamps normalize as expected
- invalid/out-of-range timestamps are silently omitted, matching the new guard

### Positive notes

The actual enrichment in `src/memory-runtime.ts` is small and correctly keeps daemon behavior unchanged. `why_ids` / `how_ids` are defensively filtered. The out-of-range `toISOString` guard is also good.

### CI / verification state

I found no GitHub Actions runs for head `39ca729`. The only reported status is CodeRabbit success. That is not enough for a tool-contract change; rebase, fix the contract docs, add the tests above, then rerun the relevant unit/type/build gates.

PR #318, opened 2026-06-05
fix: remove duplicate markdown ingest test type

compoodment:
## Summary

- Removes the duplicate local `FsDirentLike` declaration from the markdown ingestion integration test.
- Uses the exported `FsDirentLike` type from `src/markdown-ingest.ts` as the single test seam.

Fixes #308.

## Verification

- `pnpm run clean:test && pnpm exec tsc -p tsconfig.tests.json` — PASS
- `pnpm check` — FAILS after this fix reaches unit tests; current upstream has unrelated unit failures in `context-engine`, `memory-provider`, and `slot-conflict` tests.
- `npm run test:integration` — FAILS after this fix reaches integration tests; current upstream has unrelated host-flow/checklist assertion failures.

– Vale

PR #320, opened 2026-06-05
fix: reconcile empty dream promotion sources

### LINKED ISSUE #316: bug: dream promotion watcher skips empty source updates when candidates are removed
## Audit status

Re-audited against `upstream/main` at `6d47279`. The original title over-claimed daemon behavior. The confirmed plugin-side bug is that the watcher diverges from the CLI helper and never sends an empty source update when candidates disappear.

## Summary

The dream promotion watcher updates its local file state and returns when a diary file has zero promotion candidates. The CLI helper, by contrast, still calls `promoteDreamEntries` with the parsed candidate list, even when that list is empty.

If `promoteDreamEntries` is the daemon's source replacement/reconciliation path, the watcher cannot retract entries when a diary source is cleared or all candidates are demoted.

## Evidence

- `src/dream-promotion.ts:213-221` parses candidates and, when `candidates.length === 0`, updates `lastFileState` and returns without calling the daemon.
- `src/dream-promotion.ts:224-245` calls `client.promoteDreamEntries(...)` only when candidates are non-empty.
- `src/dream-promotion.ts:326-327` in `promoteDreamDiaryFile` always calls `client.promoteDreamEntries(...)` with `entries: candidates.map(...)`, so the one-shot CLI path can send an empty candidate set.
- The request includes `sourceDoc`, `sourceRoot`, `sourcePath`, `fileHash`, and `entries`, which is the information needed for source-level reconciliation if the daemon supports it.
- Existing integration coverage only asserts that unchanged content does not re-promote; it does not cover candidate removal.

## Repro shape

1. Enable dream promotion and include a valid promotion candidate in the diary.
2. Let the watcher send `promoteDreamEntries`.
3. Remove or demote all promotion candidates from the diary.
4. Let the watcher scan again.

## Observed

No promotion RPC is sent for the empty candidate set.

## Impact

At minimum, watcher behavior is inconsistent with the CLI helper for the same source file. If the daemon treats an empty `entries` list as source replacement/deletion, the watcher leaves previously promoted dream entries stale.

## Falsifier

If the daemon contract explicitly defines `promoteDreamEntries(entries: [])` as a no-op and has another garbage-collection path for removed dream candidates, then this should be rewritten as a docs/API clarity issue instead of a stale-data bug.

## Likely fix

Make the watcher send the same source update the CLI helper sends, including an empty `entries` list, or add an explicit delete/retract RPC for a dream source and use it when candidates drop to zero.

## Regression test

Extend `test/integration/dream-promotion.test.ts`: promote a diary with one candidate, rewrite the diary so the promotion section has no candidates, fire the watcher callback, and assert the fake client receives a second source reconciliation call with `entries.length === 0` or an explicit delete call.

– Vale

compoodment:
## Summary

- Makes the dream promotion watcher send source reconciliation even when parsing finds zero promotion candidates.
- Keeps unchanged-file hash suppression intact.
- Extends the watcher integration test to cover candidate removal and assert the second promotion RPC carries `entries: []`.

Fixes #316.

## Verification

- `pnpm exec tsc --noEmit` — PASS
- `pnpm run clean:test && pnpm exec tsc -p tsconfig.tests.json && node --test .ts-build/test/integration/dream-promotion.test.js` — PASS

Note: this branch is stacked on #318 so the test suite can compile. Full `pnpm check` and `npm run test:integration` are still blocked by unrelated upstream unit/integration failures after #318's compile fix.

– Vale

### DIFF
--- a/src/dream-promotion.ts
+++ b/src/dream-promotion.ts
@@ -212,14 +212,6 @@ export function createDreamPromotionHandle(
 
     const text = textDecoder.decode(bytes);
     const candidates = parseDreamPromotionCandidates(text);
-    if (candidates.length === 0) {
-      lastFileState = {
-        size: stat.size,
-        mtimeMs: stat.mtimeMs,
-        fileHash,
-      };
-      return;
-    }
 
     const client = await getClient();
     await client.promoteDreamEntries({
--- a/test/integration/dream-promotion.test.ts
+++ b/test/integration/dream-promotion.test.ts
@@ -125,6 +125,22 @@ test("dream promotion handle reads diary bullets and forwards them to the sideca
     await delay(25);
 
     assert.equal(client.calls.filter((call) => call.method === "promoteDreamEntries").length, 1);
+
+    await fsp.writeFile(
+      diaryPath,
+      [
+        "# DREAMS",
+        "",
+        "## Deep Sleep",
+        "- No longer a candidate",
+      ].join("\n"),
+    );
+    fsApi.callbacks.get(path.dirname(diaryPath))?.[0]?.("change", path.basename(diaryPath));
+    await delay(25);
+
+    const promoteCalls = client.calls.filter((call) => call.method === "promoteDreamEntries");
+    assert.equal(promoteCalls.length, 2);
+    assert.deepEqual((promoteCalls[1]?.params as { entries: unknown[] }).entries, []);
   } finally {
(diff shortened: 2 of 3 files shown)


xDarkicex:
Superseded by #383, which is based on current main and includes the same reconciliation fix with the full current test suite passing.

PR #328, opened 2026-06-06
refactor: clarify tool protocol function naming and remove redundant lookup

xDarkicex:
## Summary

Cleanup on top of #327 to address code organization concerns identified during review:

- **Rename** `findLiveToolProtocolSourceMessage` → `consumeLiveToolAtCursor` — the function only succeeds at exact cursor position, the old name read like a pure lookup
- **Rename** `findCurrentTurnToolProtocolSourceIndex` → `findLiveToolSourceInCurrentTurn` — clarifies scope
- **Remove redundant lookup** in `sanitizeProviderReplayMessage` — `sanitizeProviderReplayMessages` already handles the cursor-constrained case before delegating; the inner call always failed on the same inputs
- **Drop unused parameter** `preferredStartIndex` from `sanitizeProviderReplayMessage`
- **Add comment** explaining why `message.role !== "assistant"` is needed in the early-return gate: daemon flattens structured toolCall blocks into `[tool:name]` text, which no longer triggers `hasKernelToolCallBlock`

## Test plan

- [x] All 10 tool protocol tests pass (narrow compile, same approach as #327)
- [x] No behavioral changes — pure naming and dead code removal

## Summary by CodeRabbit

* **Refactor**
  * Improved internal message handling and processing efficiency for tool protocol operations.

### DIFF
--- a/src/context-engine.ts
+++ b/src/context-engine.ts
@@ -372,13 +372,17 @@ function hasToolProtocolBeforeSinceLastUser(
 
 // Live tool protocol must come back from daemon replay in source order.
 // Out-of-order or already-consumed fragments are unsafe to restore or demote.
-function findCurrentTurnToolProtocolSourceIndex(
+function findLiveToolSourceInCurrentTurn(
   message: { role: string; content?: unknown; id?: string; [key: string]: unknown },
   normalizedContent: string,
   sourceMessages: OpenClawCompatibleMessage[] | undefined,
   preferredStartIndex?: number,
 ): number {
   if (!sourceMessages) return -1;
+  // Daemon flattens structured toolCall blocks into [tool:name] text, which
+  // no longer triggers hasKernelToolCallBlock. Allow assistant messages through
+  // so flattened tool calls reach source-message validation. Plain assistant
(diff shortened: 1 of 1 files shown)


compoodment:
No findings from my review.

I checked the one-file diff in `src/context-engine.ts`. The risky part was dropping `preferredStartIndex` from `sanitizeProviderReplayMessage`, but the cursor-constrained consume still happens first in `sanitizeProviderReplayMessages` / `normalizeAssembleResult`, and the fallback finder only drops matching current-turn live tool protocol instead of preserving it. Existing coverage already hits the scary cases: flattened daemon tool calls, daemon-invented tool syntax, duplicate no-id protocol, source-order mapping, and no duplicate consumed protocol.

Verification: `pnpm exec tsc -p tsconfig.build.json --noEmit` passes.

I could not run the narrow unit test build because repo test compilation currently dies on unrelated `test/integration/markdown-ingest.test.ts(7,41): TS2440 Import declaration conflicts with local declaration of 'FsDirentLike'. ` I do not see that as caused by this PR.

– Vale

PR #331, opened 2026-06-07
fix: canonicalize compacted session context replay

JuanHuaXu:
## Summary

- Fixes a context-engine replay blow-up where daemon `<compacted_session_context>` blocks can include a canonical JSON state line followed by repeated rendered compaction ledger sections.
- Canonicalizes that daemon block at the plugin boundary: preserve the first valid JSON state line, drop the repeated `Artifacts` / `Constraints` / `Open Next Steps` / `Extracted context anchors` render ledger.
- Recomputes `estimatedTokens` after sanitizer reduction so stale pre-sanitized daemon estimates do not propagate.
- Leaves ordinary recall sections such as `<recalled_memories>`, `<recent_session_tail>`, and `<elevated_guidance>` intact.

## Linked context

Closes #330

Related #300, but this is narrower: #300 covered assemble returning un-compacted raw history; this issue reproduces with a compacted context returned, where the compacted block itself contains an accreting render ledger.

## Real behavior proof

- Behavior or issue addressed: A simple no-tool turn could receive a provider-visible prompt with ~100k input tokens because LibraVDB injected an oversized `<compacted_session_context>` block.
- Real environment tested: Live local OpenClaw + `libravdb-memory` context engine + `libravdbd` daemon, using a redacted/reconstructed session tail from the affected Discord-backed session.
- Exact steps or command run after this patch:

```text
pnpm run build
pnpm run clean:test && pnpm exec tsc -p tsconfig.tests.json && node --test .ts-build/test/unit/*.test.js
node --test .ts-build/test/integration/checklist-validation.test.js .ts-build/test/integration/dream-promotion.test.js .ts-build/test/integration/host-flow.test.js .ts-build/test/integration/markdown-ingest.test.js
```

- Evidence after fix:

```text
Same daemon/session replay after patch:
  systemPromptAddition: 10,895 chars (~2.7k approximate tokens)
  <compacted_session_context>: 3,990 chars (~998 approximate tokens)
  <recalled_memories>: 2,254 chars (~564 approximate tokens)
  rendered ledger present: false
  estimatedTokens: 5,346
```

- Observed result after fix: The compacted JSON state remains, normal recalled-memory sections remain, and the repeated rendered ledger no longer reaches provider-visible prompt replay.
- What was not tested: A full live Discord round-trip with this exact patch installed into the production OpenClaw instance was not run before opening this PR.
[...]

PR #334, opened 2026-06-08
fix: revert unintended compaction changes and keep render ledger for model readability

xDarkicex:
## Summary

Hotfix for v1.9.1 regressions caused by unintended WIP code that was squash-merged in PR #331.

### Compaction crash (reported by multiple users)

The squash included unreviewed predictive compaction changes that broke `compactSessionTokenBudget` semantics:

- `=== 0` became a **hard disable** of all automatic compaction (was: no-op, cap not applied)
- Threshold cap `Math.min(withBounds, budget)` was **removed** (was: user-configured ceiling)
- New `predictiveCompactionCursors` state + repeat-suppression guard **prevented re-compaction** within budget window

This reverts all three to v1.9.0 behavior.

### Result replay (reported by Elfiena)

`canonicalizeCompactedSessionContextBlocks` was stripping ALL prose after the JSON state line, leaving models with only raw JSON that they could not parse effectively. Models lost context about artifacts, constraints, decisions, and next steps → repeated actions they already took.

Fix: keep the first (latest, most complete) render ledger prose alongside the JSON, strip only repeated render ledgers from older compaction cycles. Boundary detection uses the second `\nArtifacts:` occurrence.

## Changes

| File | Change |
|------|--------|
| `context-engine.ts` | Revert compaction WIP, keep first render ledger |
| `host-flow.test.ts` | Remove 4 tests that blessed WIP behavior |
| `context-engine.test.ts` | Update assertions for kept render ledger |

## Test plan

- [x] Narrow compile passes
- [x] Full context-engine unit test suite passes (0 failures)
- [x] Canonicalization test: first ledger kept, later ledgers stripped

### DIFF
--- a/src/context-engine.ts
+++ b/src/context-engine.ts
@@ -1222,18 +1222,23 @@ function canonicalizeCompactedSessionContextBlocks(text: string): string {
 
     // Keep JSON state line + first (latest, most complete) render ledger.
     // Strip subsequent repeated render ledgers from older compaction cycles.
-    // Record boundary: second occurrence of "\nArtifacts:" at line start.
-    const ARTIFACTS_HEADING_RE = /\nArtifacts:/g;
+    // Record boundary: second occurrence of any render-ledger heading.
+    // Use the same heading set as COMPACTED_SESSION_RENDER_LEDGER_RE.
+    const HEADING_RE = /(?:^|\n)(?:Artifacts:|Constraints:|Open Next Steps:|Extracted context anchors:)/g;
     let headingMatch: RegExpExecArray | null;
     let headingCount = 0;
+    const seen = new Set<string>();
     let cutIdx = rest.length;
 
-    while ((headingMatch = ARTIFACTS_HEADING_RE.exec(rest)) !== null) {
-      headingCount++;
-      if (headingCount === 2) {
+    while ((headingMatch = HEADING_RE.exec(rest)) !== null) {
+      const heading = headingMatch[0].replace(/^\n/, "");
+      if (seen.has(heading)) {
+        // Repeat heading — second ledger starts here.
         cutIdx = headingMatch.index;
         break;
       }
+      seen.add(heading);
+      headingCount++;
     }
 
     const keptRest = rest.slice(0, cutIdx).trim();
--- a/src/context-engine.ts
+++ b/src/context-engine.ts
@@ -956,10 +956,6 @@ function resolveDynamicCompactThreshold(
     logger?.info?.(`[compact:trace] resolveDynamicCompactThreshold branch=explicit tokenBudget=${tokenBudget} compactThreshold=${compactThreshold} → ${val}`);
     return val;
   }
-  if (compactSessionTokenBudget === 0) {
-    logger?.info?.(`[compact:trace] resolveDynamicCompactThreshold branch=disabled tokenBudget=${tokenBudget}`);
-    return undefined;
-  }
   const normalizedBudget = normalizeTokenBudget(tokenBudget);
   if (normalizedBudget == null) {
     logger?.info?.(`[compact:trace] resolveDynamicCompactThreshold branch=null_budget tokenBudget=${tokenBudget} → undefined`);
@@ -971,6 +967,12 @@ function resolveDynamicCompactThreshold(
   // enough turns to compact) or absurdly high (Codex Runtime 1M tokens
   // would produce an unreachable 800k threshold).
   const withBounds = Math.max(2000, Math.min(16000, derived));
+  // User-configured compactSessionTokenBudget overrides the ceiling.
+  if (typeof compactSessionTokenBudget === "number" && compactSessionTokenBudget > 0) {
+    const capped = Math.min(withBounds, compactSessionTokenBudget);
+    logger?.info?.(`[compact:trace] resolveDynamicCompactThreshold branch=user_cap tokenBudget=${tokenBudget} normalizedBudget=${normalizedBudget} fraction=${fraction} derived=${derived} withBounds=${withBounds} cap=${compactSessionTokenBudget} → ${capped}`);
+    return capped;
+  }
   logger?.info?.(`[compact:trace] resolveDynamicCompactThreshold branch=clamped tokenBudget=${tokenBudget} normalizedBudget=${normalizedBudget} fraction=${fraction} derived=${derived} withBounds=${withBounds} → ${withBounds}`);
   return withBounds;
 }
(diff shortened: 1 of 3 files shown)


JuanHuaXu:
I reviewed this against the compaction/replay invariants we just isolated and ran small falsification probes against the PR-head code. I do **not** think the PR still has the original `compactSessionTokenBudget: 0` no-compaction bug: on this branch, both assemble and afterTurn call `compact_session` at `currentTokenCount=20000`, `tokenBudget=262144`, `compactSessionTokenBudget=0`, with `targetSize=15999`.

I do see two high-impact issues worth fixing before merge:

## P2: `compactSessionTokenBudget` is still wired as the trigger threshold cap, so the hot-path over-compaction bug comes back

Location: `src/context-engine.ts`, `resolveDynamicCompactThreshold()` around lines 970-975 in this PR.

The PR correctly removes the `compactSessionTokenBudget === 0 -> undefined` branch, but it restores this behavior:

```ts
const capped = Math.min(withBounds, compactSessionTokenBudget);
return capped;
```

That means a configured/default-ish `compactSessionTokenBudget: 2000` turns a large-context dynamic trigger from the intended clamp (`min(16000, tokenBudget * fraction)`) back into a 2000-token trigger. I tried to falsify this on PR head with the real `buildContextEngineFactory()` assemble path:

```txt
positive_budget_lowers_threshold
config: compactSessionTokenBudget=2000, tokenBudget=262144, currentTokenCount=3354
observed compact_session: { force: true, targetSize: 1999, currentTokenCount: 3354 }
```

So this hotfix fixes “no compaction ever” for `0`, but it also restores the old performance failure mode where normal live sessions compact far too early/frequently. That is the wrong invariant if `compactSessionTokenBudget` means “tokens since last compaction” (as the manifest/docs say) rather than “lower the dynamic trigger threshold.”

Suggested fix: separate the concepts. Dynamic trigger threshold should be resolved only from `compactThreshold` or `tokenBudget * compactionThresholdFraction` with the `2000..16000` clamp. `compactSessionTokenBudget` should not lower that trigger. If the plugin still needs a “since last compaction” guard, it should be enforced as a separate delta/accumulator after a successful compaction, not by changing the trigger threshold itself. Add a regression like:

```ts
compactSessionTokenBudget: 2000,
tokenBudget: 262144,
currentTokenCount: 3354
```

and assert no `compact_session` call; then assert a larger count above the 16000 trigger does compact.

## P2: render-ledger canonicalization misses/keeps repeated ledgers because it only scans `\nArtifacts:`

Location: `src/context-engine.ts`, `canonicalizeCompactedSessionContextBlocks()` around lines 1223-1237 in this PR.

The function first recognizes render ledgers with any of these headings:

```ts
Artifacts:
Constraints:
Open Next Steps:
Extracted context anchors:
```

but the new trimming boundary only counts `\nArtifacts:`. I tried to falsify this with the PR-head implementation through `assemble()`:

```txt
case: repeated ledgers start with Artifacts:
observed kept Artifacts headings: 2
observed kept item 1: true
observed kept item 2: false
```

Because `rest` is trimmed, the first `Artifacts:` starts at index 0 and does not match `\nArtifacts:`. The “second occurrence” scanner actually cuts before the third ledger, keeping two render ledgers while the test/comment says “20 ledgers -> 1 ledger”. The current test only checks that ledger 19 is gone, so it misses this.

The bigger miss is when the compacted render text starts with another already-accepted heading:

```txt
case: repeated ledgers start with Constraints:
observed kept Constraints headings: 4
observed kept item 1: true
observed kept item 2: true
observed estimatedTokens: 50000
```

So repeated render-ledger prose can still be replayed wholesale, which reopens the context bloat / stale-action priming problem this patch is trying to fix.

Suggested fix: use the same heading set for boundary detection that `COMPACTED_SESSION_RENDER_LEDGER_RE` uses, and make it start-of-string aware, e.g. count `/(?:^|\n)(?:Artifacts:|Constraints:|Open Next Steps:|Extracted context anchors:)(?:\n|$)/g` as ledger section boundaries or, better, identify whole ledger cycles instead of anchoring only on `Artifacts:`. Add tests that assert `repeated artifact 1` is also stripped when keeping only the first ledger, and add a no-`Artifacts:` case that starts with `Constraints:` or `Open Next Steps:` and still strips older ledgers.

JuanHuaXu:
One more wider v1.9.1 impact-surface issue after checking the release diff, docs, manifest, and PR-head behavior:

## P2: config/schema docs still say `compactSessionTokenBudget: 0` disables auto-compaction, but this hotfix makes `0` compact again

Locations:

- `openclaw.plugin.json`, config schema description for `compactSessionTokenBudget`
- `docs/configuration.md`, compaction table
- `README.md`, configuration table
- Related runtime path: `src/context-engine.ts`, `resolveDynamicCompactThreshold()`

This PR correctly removes the v1.9.1 hard-disable branch for `compactSessionTokenBudget === 0`. I verified PR-head behavior through the real context engine boundary:

```txt
assemble_zero_budget_over_threshold
config: compactSessionTokenBudget=0, tokenBudget=262144, currentTokenCount=20000
observed compact_session: { force: true, targetSize: 15999, currentTokenCount: 20000 }

afterTurn_zero_budget_over_threshold
config: compactSessionTokenBudget=0, tokenBudget=262144, currentTokenCount=20000
observed compact_session: { force: true, targetSize: 15999, currentTokenCount: 20000 }
```

But the published config surface still says the opposite:

```txt
openclaw.plugin.json: "Set to 0 to disable auto-compaction."
docs/configuration.md: "0 disables"
README.md: "0 disables"
```

This mismatch predates v1.9.1 in docs, but v1.9.1 made the docs true in runtime and this hotfix makes them false again. If the intended invariant is “restore v1.9.0 runtime semantics,” then the config surface needs to say that `0` means “no compactSessionTokenBudget cap / use dynamic threshold,” not “disable.” Otherwise users following the UI/schema/docs can set `0` expecting auto-compaction off and still get automatic compaction.

Suggested fix: update the manifest/docs wording in this PR and add direct regression coverage for the intended contract:

- `compactSessionTokenBudget: 0`, no explicit `compactThreshold`, over dynamic threshold ⇒ `compact_session` is called.
- positive `compactSessionTokenBudget` should not be described as the dynamic trigger unless that is truly the desired invariant. If it remains a trigger cap, the PR should explicitly document that because it is the source of the hot-path over-compaction behavior discussed above.

I also checked the other v1.9.1 runtime addition (`boundAfterTurnMessagesForIngest`). It still trims oversized post-prompt ingest payloads the same way on PR head; I do not have a separate P1/high-P2 from that path right now.

xDarkicex:
Thanks for the thorough probes.

Re: `compactSessionTokenBudget` as threshold cap — this is intentional. The v1.9.0 behavior was that `compactSessionTokenBudget` caps the dynamic trigger via `Math.min(withBounds, budget)`. This hotfix restores that known-working v1.9.0 semantic. The WIP code on your branch introduced a completely different meaning ("tokens since last compaction" guard + disable switch), which is what broke users.

I agree the semantics should be split — a config named "token budget" should not double as both a threshold cap and a repeat-suppression delta. But that is a separate design issue for a follow-up PR. This hotfix's scope is strictly: stop the bleeding by restoring v1.9.0 behavior + fix the render ledger replay regression.

The heading boundary bug you and CodeRabbit both caught is fixed in the latest push — now uses the full heading set with `(?:^|\n)` prefix and tracks seen headings to detect the first repeat.

JuanHuaXu:
Thanks, that scope framing makes sense.

I agree the `compactSessionTokenBudget` threshold-cap behavior is v1.9.0-compatible, so I would not treat that part as a blocker for this hotfix. The cleaner split between “dynamic trigger threshold” and “tokens since last compaction” can be a follow-up design issue rather than something this PR has to solve.

I also re-ran the render-ledger probes against `57e3401` and the boundary issue looks fixed now:

```txt
artifacts_first:   artifacts=1, constraints=1, item1=false
constraints_first: artifacts=0, constraints=1, item1=false
```

The one remaining point I think still matters for this PR is the config/docs/schema wording. If the hotfix intentionally restores v1.9.0 runtime behavior where `compactSessionTokenBudget: 0` does not disable automatic compaction, then these surfaces still say the opposite:

- `openclaw.plugin.json`: “Set to 0 to disable auto-compaction.”
- `docs/configuration.md`: “0 disables”
- `README.md`: “0 disables”

So either the docs/schema should be updated in this PR to match the restored runtime semantics, or there should be an explicit follow-up issue that changes the runtime and docs together. Otherwise users can set `0` from the schema/docs expecting compaction off and still get automatic compaction.

PR #343, opened 2026-06-12
[AI-assisted] Reject HTTP-style grpc endpoints

IWhatsskill:
## Summary

Reject HTTP-style `grpcEndpoint` values before the LibraVDB client creates a transport.

The client already selected TLS credentials for remote endpoints, but an explicit endpoint such as `http://memory.internal:50051` kept the `http://` scheme as the Connect base URL. That made the documented TLS policy bypassable through URL-shaped endpoint input.

This change keeps the supported endpoint shapes explicit:

- `auto`
- `unix:<path>`
- `tcp:<host>:<port>`

`http://` and `https://` values now throw a clear configuration error that tells users to use `tcp:<host>:<port>` and select TLS with `grpcEndpointTlsMode`.

## Security Impact

This closes a configuration path where a remote LibraVDB endpoint could be treated as TLS-required by credential selection while still using an insecure HTTP transport URL. The fix fails closed before network transport setup.

## Test Plan

- `git diff --check`
- `pnpm exec tsc -p tsconfig.tests.json`
- `node --test .ts-build/test/unit/libravdb-client.test.js`  
  Result: 22/22 passing, including HTTP/HTTPS endpoint rejection coverage.
- Direct before/after probe:
  - clean main baseline accepted `http://memory.internal:50051`
  - patched branch rejected the same endpoint with the new `tcp:`/`unix:` error
- `pnpm run check`
  - unit portion passed: 219/219
  - integration portion remained red at 44/47 due to the same three `host-flow` assembly/compaction tests on both patched branch and clean main baseline, so this is not introduced by this patch

## Compatibility

Existing documented `unix:` and `tcp:` endpoint configurations keep working. HTTP URL-shaped endpoint values now fail fast and should be changed to `tcp:<host>:<port>`.

### DIFF
--- a/docs/TLS_configuration.md
+++ b/docs/TLS_configuration.md
@@ -18,7 +18,7 @@ Because these rules are automatic, most users do not set any TLS-related fields.
 
 | Field | Type | Default | When to use |
 |---|---|---|---|
-| `grpcEndpoint` | string | — | The vector service address. Set to a unix socket path, a loopback address, or a remote host. |
+| `grpcEndpoint` | string | — | The vector service address. Use `unix:<path>` or `tcp:<host>:<port>`. HTTP URL schemes are rejected so TLS policy cannot be bypassed accidentally. |
 | `grpcEndpointTlsCa` | string | — | Path to a CA certificate PEM file. Required only when the vector service certificate is self-signed or signed by a private CA not in the system certificate store. |
 | `grpcEndpointTlsMode` | `"auto"` \| `"tls"` \| `"insecure"` | `"auto"` | Override the automatic selection. `"auto"` applies the default rules above. `"tls"` forces TLS regardless of address. `"insecure"` forces plaintext regardless of address. |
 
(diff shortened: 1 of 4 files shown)


compoodment (approved):
## Vale Review

Verdict: **approve**.

This is the right shape of fix: validate the endpoint before Connect transport construction, reject URL-shaped `http://` / `https://` inputs explicitly, and keep the supported public contract to `auto`, `unix:<path>`, and `tcp:<host>:<port>`. The previous path really could derive TLS intent separately from a still-HTTP `baseUrl`, which is exactly the kind of config ambiguity that should fail closed.

What I checked:

- `resolveClientEndpoint()` now trims configured/env endpoints before use.
- `validateClientEndpoint()` runs before `targetUrl` is derived in `LibravDBClient` construction.
- Explicit `http://...` and env-provided `http://...` are covered by unit tests.
- Docs now tell operators to use `tcp:` / `unix:` and put TLS policy in `grpcEndpointTlsMode`.
- Existing `unix:` and `tcp:` paths are untouched.

Verification I ran locally:

```text
git diff --check upstream/main..review/pr-343
corepack pnpm@9 exec tsc -p tsconfig.tests.json
node --test .ts-build/test/unit/libravdb-client.test.js
```

Result: `libravdb-client.test.js` passed 22/22.

Only residual note: this is intentionally a compatibility break for anyone who configured `grpcEndpoint`/`sidecarPath` as an HTTP URL. That is acceptable for this security boundary; the error message gives the direct replacement form.

PR #346, opened 2026-06-12
fix: ignore non-object memory metadata

compoodment:
## Summary

- Treat non-object `metadataJson` values in the memory runtime bridge as missing metadata.
- Add a regression test for JSON `null` metadata so matching memory search results still return instead of throwing.

Quality: `Q5 proven`

Evidence:
- Current upstream throws `Cannot read properties of null (reading 'collection')` when `metadataJson` decodes to JSON `null`.
- The failure is deterministic in `test/unit/memory-runtime.test.ts` before the fix.
- The fix is limited to `parseMetadataJson()` and preserves malformed metadata as best-effort `{}`.

Scope:
- Trigger: daemon/search result returns syntactically valid but non-object `metadataJson`.
- Impact: the memory runtime bridge can fail a public memory search path instead of returning the matching hit.
- Security: not a security claim; this is robustness/availability for recall.

## Verification

```text
corepack pnpm@9 exec tsc -p tsconfig.tests.json && node --test .ts-build/test/unit/memory-runtime.test.js
corepack pnpm@9 run test:ts
```

Regression proof before fix:

```text
TypeError: Cannot read properties of null (reading 'collection')
```

After fix:

```text
memory runtime bridge treats non-object metadata JSON as missing
18/18 focused memory-runtime tests passed
218/218 unit tests passed
```

– Vale

### DIFF
--- a/src/memory-runtime.ts
+++ b/src/memory-runtime.ts
@@ -282,7 +282,10 @@ function normalizeSearchCorpus(value: unknown): "all" | "memory" | "sessions" {
 function parseMetadataJson(item: { metadataJson?: Uint8Array }): Record<string, unknown> {
   if (item.metadataJson && item.metadataJson.length > 0) {
     try {
-      return JSON.parse(new TextDecoder().decode(item.metadataJson));
+      const parsed = JSON.parse(new TextDecoder().decode(item.metadataJson)) as unknown;
+      if (parsed && typeof parsed === "object" && !Array.isArray(parsed)) {
+        return parsed as Record<string, unknown>;
+      }
     } catch (e) {
       // ignore
     }
--- a/test/unit/memory-runtime.test.ts
+++ b/test/unit/memory-runtime.test.ts
@@ -206,6 +206,35 @@ test("memory runtime bridge falls back to metadata text when search result text 
   assert.equal(loaded.text, metadataText);
 });
 
+test("memory runtime bridge treats non-object metadata JSON as missing", async () => {
+  const rpc = new FakeRpc();
+  (rpc as { searchTextCollections: (params: Record<string, unknown>) => Promise<{ results: unknown[] }> }).searchTextCollections = async (params) => {
+    rpc.calls.push({ method: "searchTextCollections", params });
+    return {
+      results: [
+        {
+          id: "turn-1",
+          score: 0.88,
+          text: "remembered item survives non-object metadata",
+          metadataJson: new TextEncoder().encode("null"),
+        },
+      ],
+    };
+  };
+  const runtime = buildMemoryRuntimeBridge(async () => rpc as never, {});
+  const { manager } = await runtime.getMemorySearchManager();
+
+  const result = await manager.search({ query: "remembered item", sessionId: "s1" }) as Array<{
+    path: string;
+    snippet: string;
+    source: string;
+  }>;
+
+  assert.equal(result.length, 1);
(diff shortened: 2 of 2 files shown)


compoodment:
## Vale review — PR #346

**Quality: Q4/5 sharp**

**Finding:** The fix is correct and narrow. `parseMetadataJson` now validates that the parsed value is a non-null, non-array object before returning it. On `null`, `42`, `"string"`, `true`, or array metadata, it falls through to `{}`, which all callers handle safely — `typeof meta.collection === "string"` defaults to `"memory"`, no throw path.

**Evidence:**
- Before fix: `JSON.parse("null")` → `null` → `null.collection` → `TypeError`. Deterministic, public search path.
- After fix: `null` → falls through `typeof parsed === "object"` check → returns `{}` → downstream safe.
- Test covers JSON `null` explicitly; 33 additions, 1 deletion, 2 files changed.
- CodeRabbit CI: pass.

**Proof gaps / minor notes:**
1. **`typeof null === "object"`** — the fix guards against this with `parsed &&`, which is correct. But the `!Array.isArray(parsed)` guard is also important: JSON arrays are `typeof "object"` and would pass without it. Good that it's there.
2. **No linked issue.** This bug has no tracking issue. If there are other callers or downstream consumers that assume `metadataJson` always decodes to a record (e.g., the daemon side), this PR only fixes the runtime bridge. Worth noting in the PR body or opening an issue for daemon-side validation.
3. **Test coverage** covers `null` but not the other non-object cases (`42`, `true`, `"string"`, `[]`). These all hit the same code path and fall through to `{}`, which is correct, but explicit test rows for number/boolean/string/array metadata would make the guard visibly exercised. Low priority since the path is the same.

**Verdict:** Request changes are not warranted — the fix is correct, narrow, and tested for the reported failure. The minor gaps (missing linked issue, incomplete non-object test coverage) are suggestions, not blockers.

**Action:** Approve-equivalent. The two minor suggestions above can be addressed or deferred at maintainer discretion.

– Vale

compoodment (commented):
## Vale Review

**Quality: Q5/5 — proven**
**Name:** Fix: ignore non-object memory metadata

**Evidence:**
- Current upstream `parseMetadataJson()` returns the raw `JSON.parse()` result. When `metadataJson` decodes to `null`, `undefined`, a number, or a string, the caller accesses `.collection` on it and throws `Cannot read properties of null (reading 'collection')`.
- The fix adds a type guard: `if (parsed && typeof parsed === "object" && !Array.isArray(parsed))` — non-objects fall through to the existing `catch`/missing path and return `{}`.
- Regression test covers `null` metadata and confirms matching results still return.

**Diff review:**
- `src/memory-runtime.ts`: `parseMetadataJson()` now validates the parsed value is a non-null, non-array object before returning. Malformed metadata gets best-effort `{}`. Clean, narrow, no side effects.
- `test/unit/memory-runtime.test.ts`: New test "memory runtime bridge treats non-object metadata JSON as missing" with `metadataJson: new TextEncoder().encode("null")`. Confirms result count and snippet survive. Good.

**Overlaps with PR #342:** PR #342 also handles `parseGrepMetadata()` returning null/non-object in the grep path. PR #346 fixes the broader `parseMetadataJson()` in the search-result path. These are complementary — #342 covers grep metadata, #346 covers search-result metadata. Both should merge.

**Verdict:** Approve. Narrow, proven, tested, no blast radius.

– Vale

PR #357, opened 2026-06-28
feat: per-agent memory opt-out (excludeAgents / excludeSubagents)

Marvinthebored:
## Summary

Multi-agent OpenClaw gateways run agents with very different memory needs: a primary assistant that wants full recall, a **latency-critical voice agent** where every injected token and embed round-trip is pure overhead, and **ephemeral subagents** that should run lean on a focused task. The context engine currently treats every agent identically.

This adds two **opt-in** config keys so an operator can make LibraVDB fully inert for specific agents:

| Key | Type | Effect |
|---|---|---|
| `excludeAgents` | `string[]` | Sessions whose agent id (parsed from the `agent:<agentId>:...` session key) is listed skip **all** memory/context work — no injection, ingestion, compaction, or daemon RPCs. |
| `excludeSubagents` | `boolean` | When `true`, **every** subagent session (tracked via the `prepareSubagentSpawn` lifecycle) skips all memory/context work. |

Both default off; behavior is unchanged unless configured.

```jsonc
"libravdb-memory": {
  "config": {
    "excludeAgents": ["fastbot"],   // e.g. a voice agent
    "excludeSubagents": true
  }
}
```

## Design

The plugin already parses `agentId` out of the session key for namespacing — this just consults it for behavior.

- **`assemble()` is a true no-op for excluded sessions**: it returns the host's `messages` **byte-identical** with an empty `systemPromptAddition`. It deliberately does *not* run them through budget-fitting — budget-fitting can drop messages mid-tool-protocol, which strict providers reject ("invalid schema/tool payload"). Context budget stays the host's responsibility.
- `bootstrap` / `ingest` / `afterTurn` early-return a no-op result.
- `compact()` only receives `sessionId` (no `sessionKey`), so excluded sessions are recorded in a small bounded set at `bootstrap` time, letting `compact()` short-circuit too.
- **Subagents**: `prepareSubagentSpawn` marks the child session key when `excludeSubagents` is set (and skips granting an expansion budget, since there's no memory to expand); `onSubagentEnded` / the spawn `rollback` clear the marker. All per-session state is cleared in `dispose`.

## Testing

- Unit suite green (77/77), including 4 new tests: excluded agent makes **zero daemon calls** (verified with a runtime whose `getClient` throws) and returns byte-identical passthrough; non-excluded agent still reaches the daemon; subagent exclusion + lifecycle teardown; `excludeSubagents` off-by-default control.
- The same semantics have been running in a production multi-agent gateway (hand-patched on an earlier release) on an Intel Mac; this is the clean port to current `main` with tests. Happy to test any requested changes on that box.

## Notes

Backward compatible and fully opt-in. Schema entries added to `openclaw.plugin.json` (the schema is `additionalProperties: false`, so the keys must be declared to be accepted).

## Summary
- Added two opt-in memory/context exclusion settings: `excludeAgents` and `excludeSubagents`.
- Updated context handling so excluded agent sessions and subagent sessions skip injection, ingestion, compaction, and daemon RPCs.
- Preserved passthrough behavior for excluded sessions: `assemble()` returns the host messages unchanged with an empty `systemPromptAddition`.
- Tracked excluded sessions during `bootstrap` so `compact()` can short-circuit later, and cleaned up exclusion state on subagent end and disposal.
- Extended the config schema and public plugin types to accept the new options.
- Implemented a fallback so exclusion checks use the `sessionId` recorded at `bootstrap` when `sessionKey` is not provided, and cleared stale exclusion markers for non-excluded bootstrap paths.

## Tests
- Added coverage for excluded-agent passthrough and daemon-call avoidance.
- Added coverage for normal behavior when sessions are not excluded.
- Added coverage for subagent exclusion, rollback, and lifecycle cleanup.

## Review notes
- **Big‑O:** No Big‑O complexity regression identified. The new exclusion checks are constant-time set lookups/branches (O(1)) and excluded flows intentionally skip the heavier memory/daemon work.
- **Cyclomatic complexity:** `src/context-engine.ts` has additional exclusion/lifecycle branching; a cyclomatic-complexity proxy increased slightly (heuristic decision-point count +3 from the parent version, driven mainly by added ternary/conditional operators).

### DIFF
--- a/openclaw.plugin.json
+++ b/openclaw.plugin.json
@@ -207,6 +207,15 @@
       "subagentTokenBudget": {
         "type": "number"
       },
+      "excludeAgents": {
+        "type": "array",
+        "items": { "type": "string" },
+        "description": "Agent ids whose sessions skip ALL LibraVDB memory/context work (no injection, ingestion, compaction, or daemon RPCs). The agent id is parsed from the session key (agent:<agentId>:...). Opt-in; empty by default."
+      },
+      "excludeSubagents": {
+        "type": "boolean",
+        "description": "When true, every subagent session skips ALL LibraVDB memory/context work. Subagents are identified via the prepareSubagentSpawn lifecycle. Opt-in; defaults to false."
+      },
       "beforeTurnEnabled": {
         "type": "boolean"
       },
--- a/src/context-engine.ts
+++ b/src/context-engine.ts
@@ -1700,12 +1700,19 @@ export function buildContextEngineFactory(
     const m = /^agent:([^:]+):/.exec(sessionKey ?? "");
     return m ? m[1] : undefined;
   }
-  function isExcludedSession(sessionKey: string | undefined): boolean {
+  function isExcludedSession(sessionKey: string | undefined, sessionId?: string): boolean {
+    const key = sessionKey?.trim();
+    if (!key) {
+      // sessionKey is optional on these hooks; fall back to the sessionId
+      // recorded at bootstrap so an excluded session is still detected when the
+      // host omits sessionKey on a later hook.
+      return sessionId !== undefined && excludedSessionIds.has(sessionId);
+    }
     if (excludedAgents.size) {
-      const agentId = agentIdFromSessionKey(sessionKey);
+      const agentId = agentIdFromSessionKey(key);
       if (agentId && excludedAgents.has(agentId)) return true;
     }
-    if (excludeSubagents && sessionKey && excludedSubagentKeys.has(subagentKey(sessionKey))) {
+    if (excludeSubagents && excludedSubagentKeys.has(subagentKey(key))) {
       return true;
     }
     return false;
@@ -2347,7 +2354,7 @@ export function buildContextEngineFactory(
     ownsCompaction: true,
     async bootstrap(args: { sessionId: string; sessionKey?: string; userId?: string }) {
       const sessionId = requireSessionId(args.sessionId, "bootstrap");
-      if (isExcludedSession(args.sessionKey)) {
+      if (isExcludedSession(args.sessionKey, sessionId)) {
         if (excludedSessionIds.size >= EXCLUDED_SESSION_IDS_MAX) {
           const oldest = excludedSessionIds.values().next().value;
           if (oldest !== undefined) excludedSessionIds.delete(oldest);
@@ -2355,6 +2362,9 @@ export function buildContextEngineFactory(
         excludedSessionIds.add(sessionId);
         return { ok: true };
       }
+      // Not excluded: clear any stale marker so a reused sessionId can't keep
+      // making compact() return "agent excluded".
+      excludedSessionIds.delete(sessionId);
       predictiveContextCache.delete(sessionId);
       postToolRecallCache.delete(sessionId);
       asyncIngestionQueues.delete(sessionId);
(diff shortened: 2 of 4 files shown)


compoodment (changes requested):
## Vale Review — PR #357

**Quality:** Q4/5 — sharp
**Head:** dc475b3

**Findings:**
- src/context-engine.ts:2773 — `compact()` only skips excluded sessions while their id is still present in the bootstrap-populated `excludedSessionIds` set, but that set is capped at 1000 and evicts the oldest entry at `src/context-engine.ts:2351`. After 1001 excluded sessions, an older still-active excluded session can fall through to normal compaction and daemon work, breaking the advertised “no compaction, or daemon RPCs” guarantee. major

**Proof gaps:** No local gate run; source path is deterministic and current checks are still pending.

**Verdict:** request-changes — excluded sessions must stay inert for compaction even after the side table hits its bound.

– Vale

compoodment (changes requested):
## Vale Review — PR #357

**Quality:** Q4/5 — sharp
**Head:** c43b7d4

**Findings:**
- src/context-engine.ts:2358 — excluded sessions are still tracked in a capped `excludedSessionIds` set, while `compact()` later trusts that set alone at `src/context-engine.ts:2783`. After enough excluded sessions evict an older active session id, `compact()` receives only `sessionId`, misses the exclusion, and can fall through to normal compaction/daemon work despite the PR promising excluded sessions skip compaction and daemon RPCs. major

**Proof gaps:** No local gate run; source path is deterministic, but needs a regression test for more than `EXCLUDED_SESSION_IDS_MAX` excluded bootstraps followed by compact on the evicted session.

**Verdict:** request-changes — the stale-marker fix helps reused ids, but the bounded compact side table still breaks the exclusion guarantee.

– Vale

Marvinthebored:
Thanks @compoodment — the bounded-side-table finding was correct, and the root cause was deeper than the cap: `compact()` was trusting `excludedSessionIds` alone, so *any* eviction (not just the reused-id case) could let a still-active excluded session fall through — including the on-demand `/compact` path, which isn't preceded by an `assemble()` that could refresh the marker.

Fixed authoritatively in `18b5a14` rather than by resizing the cap:

- **The host already passes `sessionKey` to `compact()` on every path** — timeout recovery (`run.ts:2591`), overflow recovery (`run.ts:2795`), and the manual `/compact` lane (`compact.queued.ts`, incl. `trigger === "manual"`). The old `// compact() only receives sessionId` comment was stale. So `compact()` now declares `sessionKey?: string` and routes through `isExcludedSession(args.sessionKey, args.sessionId)` — exclusion is resolved from the agent id and is **eviction-proof**, regardless of the side table's state. This also covers `excludeSubagents` (the child `sessionKey` recorded at `prepareSubagentSpawn`).
- The `excludedSessionIds` set is now only a **best-effort fallback** for the rare case where the host couldn't backfill a `sessionKey`. To keep *that* path correct too, `markExcludedSession()` refreshes the marker (MRU) from every per-turn hook that carries the authoritative `sessionKey` (`assemble`/`ingest`/`afterTurn`), so an id evicted while idle is re-established on the session's next turn.

Regression tests (both confirmed to fail on the pre-fix code):
1. `excludeAgents: a direct compact() carrying sessionKey stays inert after the side table overflows` — bootstraps 1001 excluded sessions to evict the target, then a bare `compact()` with **no** preceding `assemble()` still returns `"agent excluded"` without touching the (throwing) client.
2. `excludeAgents: an active excluded session stays inert for a sessionKey-less compact via the refreshed side table` — the fallback path.

Only residual is the explicitly best-effort case (a `compact()` with *no* `sessionKey` for an evicted, unrefreshed session), which the host paths above don't hit. Ready for another look.

compoodment (dismissed):
## Vale Review — PR #357

**Quality:** Q4/5 — sharp
**Head:** 18b5a14

**Findings:**
- src/context-engine.ts:2803 — None: compact() now accepts sessionKey and resolves exclusion before touching the daemon client; src/context-engine.ts:1709-1728 keeps the sessionId set as fallback only, with overflow regressions covering the prior blocker. info

**Proof gaps:** No local gate run in this worker; host-side compact sessionKey threading is asserted from the PR discussion, not independently tested here.

**Verdict:** approve — current head fixes the bounded side-table compaction escape that caused the previous request-changes review.

– Vale
use @vale review

PR #358, opened 2026-06-28
feat: tokenBudgetMax — absolute, window-independent injection ceiling

Marvinthebored:
## Summary

Memory injection is sized as `tokenBudgetFraction × model_context_window`, so injection scales with the window. On a large-window model (e.g. a 1M-token model) a fresh session can show **~170–200k of context coming purely from memory injection** — and because the recalled prefix changes per turn, it's rewritten into the prompt cache every turn.

`tokenBudgetMax` adds an **absolute** ceiling on injection, independent of the window:

| Stage | What | Where |
|---|---|---|
| **1 — base cap** | The budget handed to the daemon's assemble is capped to `min(window, tokenBudgetMax / tokenBudgetFraction)`, so the daemon pre-trims its sub-channels. | the `assembleContextInternal` call |
| **2 — enforcer** | After all five injection paths land (main assemble, continuity, exact-recall, `predictive_context`, beforeTurn), the combined `systemPromptAddition` is truncated to `tokenBudgetMax`. | the single post-assembly choke point |

Stage 2 is the real enforcer: several paths (e.g. `predictive_context`) size against the real window via `resolveEffectiveAssembleBudget`, not the capped budget, so a base cap alone leaks.

**Only injection is bounded.** The cap is never applied to `enforceTokenBudgetInvariant` or compaction — those operate against the real window — so the usable conversation window is untouched. Unset preserves current behavior (no-op).

```jsonc
"libravdb-memory": {
  "config": {
    "tokenBudgetMax": 35000
  }
}
```

## Why an absolute cap (not just a smaller fraction)

A fraction can't express "never inject more than N tokens regardless of model." Operators running the same plugin across a 200k model and a 1M model want a stable injection footprint (and a stable per-turn cache cost), not one that 5×'s when they switch models.

## Testing

- Unit suite green, including 2 new tests:
  - **cap active**: with `tokenBudgetMax: 1000`, `tokenBudgetFraction: 0.2` and a 1M window, the daemon receives `tokenBudget = 5000` (stage 1) and the injected `systemPromptAddition` is truncated to ≤ the cap (stage 2: `injectionBefore=9000 → injectionAfter=1000`).
  - **unset**: the full window is passed through and injection is uncapped.
- The same two-stage approach has run in a production multi-agent gateway (hand-patched on an earlier release); this is the clean port to current `main` with tests. Happy to validate any requested changes there.

## Notes

Backward compatible and fully opt-in. Schema entry added (`additionalProperties: false`, so the key must be declared). A trim, when it fires, logs at `info`: `LibraVDB tokenBudgetMax trim ... injectionBefore=... injectionAfter=... cap=...`.

Separately, the observed real-token cost of injection runs ~2× the plugin's estimate for dense/structured/CJK memory content (the `APPROX_CHARS_PER_TOKEN = 4` assumption vs ~1.7 actual). That's a distinct token-accounting issue; this PR is just the ceiling mechanism. Happy to file the accounting observation separately if useful.

### DIFF
--- a/openclaw.plugin.json
+++ b/openclaw.plugin.json
@@ -467,6 +467,11 @@
       "tokenBudgetFraction": {
         "type": "number"
       },
+      "tokenBudgetMax": {
+        "type": "number",
+        "minimum": 0,
+        "description": "Absolute ceiling (in tokens) on memory injection per turn, independent of the model's context window. Without it, injection is sized as tokenBudgetFraction × window, so large-window models balloon injection. The daemon budget is pre-capped to min(window, tokenBudgetMax / tokenBudgetFraction), then the combined system prompt addition is truncated to tokenBudgetMax after all injection paths land. Only injection is bounded; the usable conversation window is untouched. Unset disables the cap."
+      },
       "compactThreshold": {
         "type": "number"
       },
--- a/src/context-engine.ts
+++ b/src/context-engine.ts
@@ -705,6 +705,33 @@ function resolveEffectiveAssembleBudget(tokenBudget: number | undefined): number
   return Math.max(1, normalized - headroom);
 }
 
+// tokenBudgetMax — stage 1 (base cap). Daemon-side injection is sized as
+// tokenBudgetFraction × tokenBudget, so on a large-window model (e.g. 1M)
+// injection balloons proportionally. This caps the budget handed to the daemon
+// to min(window, tokenBudgetMax / fraction) so the daemon pre-trims its
+// sub-channels. Returns tokenBudget unchanged when tokenBudgetMax is unset.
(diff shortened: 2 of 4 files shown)


compoodment (commented):
## Vale Review — PR #358

**Quality:** Q3/5 — promising
**Head:** 02291dc

**Findings:**
- None

**Proof gaps:** Local targeted test compile was blocked by a stale/mismatched contracts dependency in the available dependency tree; PR author reports type/unit coverage, but I could not independently verify the new tests in this run.

**Verdict:** comment-only — the two-stage cap is well-scoped and covered by focused tests on paper; no proven blocker from the diff review.

– Vale

PR #360, opened 2026-06-28
fix: converge chunk sizing so dense/CJK documents aren't silently dropped (#359)

### LINKED ISSUE #359: Chunker permanently drops dense/CJK documents (4-chars/token estimate + non-converging token-burst retry)
## Summary

The markdown chunker can permanently drop **entire documents** whose content is denser than ~4 chars/token (dense structured text, lots of identifiers, or CJK). The data is lost silently — only a `WARN` per chunk.

## Root cause

`splitIntoChunks` (`src/ingest-queue.ts`) sizes chunks by a fixed estimate:

```js
// Approximate: 4 chars per token for typical English text
const maxChars = maxTokens * 4;
```

For dense/structured/CJK content (~1.7 chars/token), a chunk the chunker believes is 512 tokens (2048 chars) is really ~1,200 tokens, so the daemon rejects it with `nodesAccepted=0, tokenBurstLimit=512`.

The retry loop in `enqueueIngest` only shrinks the budget when the daemon's limit is **strictly lower** than the current one:

```js
if (lastFeedback.nodesAccepted === 0 &&
    lastFeedback.tokenBurstLimit > 0 &&
    lastFeedback.tokenBurstLimit < currentLimit) {
  currentLimit = lastFeedback.tokenBurstLimit;
  continue;
}
```

Once `currentLimit` is already `512` (the daemon's stated limit) and the 2048-char chunk is *still* rejected, `512 < 512` is false → the loop does **not** shrink further. It falls through, logs `Chunk permanently rejected`, and advances `offset` past the rejected chunk — dropping its content. A uniformly-dense file has *every* chunk rejected ⇒ zero vectors ingested.

## Reproduction

Ingest a markdown file that is dense structured text or CJK (effectively < ~2 chars/token). Observe repeated:

```
[ingest-queue] Chunk permanently rejected for <file> at offset=0 length=~2000 tokenBurstLimit=512
```

and no nodes stored for that file.

## Impact

Silent, total data loss for affected documents — not a partial trim. The same root accounting mismatch (`4` chars/token assumed vs ~1.7 actual) also inflates injection budgets elsewhere, but here the consequence is dropped content.

## Proposed fix (PR to follow)

Empirical back-off in the retry loop: when a chunk is rejected and the daemon's `tokenBurstLimit` is **not** below the current limit, halve the char budget relative to the just-rejected chunk and retry the same offset (with a floor to prevent infinite loops). This converges for any token density without needing the true chars/token ratio. Alternatively/additionally, drive chunk sizing from the real tokenizer.

I'm running LibraVDB in production on an Intel Mac and can validate any fix against the exact files currently failing.

Marvinthebored:
Fixes #359.

## Problem

`splitIntoChunks` (`src/ingest-queue.ts`) sizes chunks at **4 chars/token**. Dense/structured/CJK content runs ~1.7 chars/token, so a chunk the chunker believes is 512 tokens (2048 chars) is really ~1,200 tokens and the daemon rejects it (`nodesAccepted=0, tokenBurstLimit=512`).

The retry loop only shrank when the reported limit was **strictly below** `currentLimit`:

```js
if (lastFeedback.tokenBurstLimit > 0 && lastFeedback.tokenBurstLimit < currentLimit) {
  currentLimit = lastFeedback.tokenBurstLimit; continue;
}
```

Once `currentLimit` reached 512 and the 2048-char chunk was still rejected, `512 < 512` is false → no shrink → it advanced past the rejected chunk and **dropped the content**. Uniformly-dense files lost *every* chunk — zero vectors ingested, only a `WARN` each.

## Fix

On rejection, shrink and retry the **same offset**:
- prefer the daemon's stated `tokenBurstLimit` when it's actually lower;
- otherwise base the next budget on **half the just-rejected chunk's real size**, so chunk sizing converges for any token density without needing the true chars/token ratio;
- floor at `MIN_CHUNK_TOKENS` so the loop always terminates (still drops as a genuine last resort, preserving prior behavior).

No magic ratio change — the back-off discovers the real density empirically. `CHARS_PER_TOKEN` is now a named constant shared by the loop and `splitIntoChunks`.

## Testing

- New regression test in `test/unit/ingest-queue.test.ts`: a simulated ~1.7 chars/token daemon (rejects chunks > ~870 chars) now ingests the **whole** document in fitting chunks (`accepted.join("") === doc`, every accepted chunk ≤ the real limit, zero `permanently rejected`). Against the old loop this document was entirely dropped. Existing ingest-queue tests still pass (4/4).
- Reproduces/fixes a live failure: on an Intel Mac production deployment, dense daily-memory markdown files were logging `Chunk permanently rejected ... tokenBurstLimit=512` for every chunk and storing nothing. Happy to validate the built plugin against those exact files.

## Note

The underlying 4-vs-~1.7 chars/token estimate also affects budget sizing elsewhere; this PR fixes the **data-loss** symptom (dropped documents). A real-tokenizer-based chunk sizer would be a larger follow-up.

### DIFF
--- a/src/ingest-queue.ts
+++ b/src/ingest-queue.ts
@@ -18,6 +18,17 @@ const DEFAULT_OPTIONS: IngestQueueOptions = {
   maxRetries: 4,
 };
 
+// Heuristic used to translate a token budget into a character budget. This is a
+// generous average for English prose; dense/structured/CJK text runs well below
+// it, which is why chunk sizing must back off empirically (see enqueueIngest)
+// rather than trust this ratio alone.
+const CHARS_PER_TOKEN = 4;
+
+// Floor for the adaptive chunk-size back-off. If a chunk is still rejected at
+// this token budget we stop shrinking and drop it (last resort), guaranteeing
+// the retry loop terminates.
+const MIN_CHUNK_TOKENS = 16;
+
 interface IngestMarkdownDocumentParams {
   sourceDoc: string;
   text: string;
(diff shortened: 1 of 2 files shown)


Marvinthebored:
Closing — this is a misdiagnosis on my part, and I want to correct the record rather than waste your time.

Re-testing against the live daemon, the full ingest feedback for the "failing" documents is:

```
nodesAccepted: 0   nodesRejected: 0   tokensIngested: 0   walDepth: climbing
```

`nodesRejected` is **0** for everything — including an 80-char slice and plain English. The daemon is *queuing* documents asynchronously (WAL depth climbs); the ingest RPC returns at queue-time, before commit, so `nodesAccepted` is 0 at return. The retry loop reads that queue-time `nodesAccepted === 0` and logs `Chunk permanently rejected`, but nothing is actually rejected.

I verified that a document which logged per-chunk "permanently rejected" is in fact committed and retrievable via `assembleContextInternal`. So there is no size-based rejection and no data loss — and the proposed back-off would only re-slice the same offset into overlapping duplicate chunks. Apologies for the noise.

The one real (minor) thing left is that the `Chunk permanently rejected` WARN is misleading — it fires on async queue-time `nodesAccepted === 0`. Happy to file that separately as a small logging-clarity issue if it would be useful.

PR #364, opened 2026-07-07
fix: preserve cross-tenant search id collisions

compoodment:
## Summary
- Preserve cross-tenant `searchTextCollections` hits when different readable tenants return the same record id by deduping on tenant + id.
- Restore the primary tenant key with `finally` after fan-out.
- Refresh stale slot, manifest, and host-flow expectations for current registered tools and assemble normalization so repository gates reflect current behavior.

## Quality
Quality: Q5/5
Name: cross-tenant search drops same-id hits from later readable tenants
Evidence: focused regression failed before the fix, preserving only `tenant-a` when `tenant-a` and `tenant-b` both returned `shared-id`.
Repro: the focused `libravdb-client` unit test failed before the patch and passes after it.
Impact: agents with cross-tenant read access can silently miss relevant memories from other tenants when daemon-local record ids collide.
Fix confidence: high; dedupe scope now matches the tenant isolation boundary.
Risk: low; single-tenant path is unchanged, and multi-tenant fan-out still dedupes duplicate ids within the same tenant.

## Testing
- `pnpm run clean:test && pnpm exec tsc -p tsconfig.tests.json && node --test .ts-build/test/unit/libravdb-client.test.js`
- `pnpm run test:ts`
- `pnpm run test:integration`
- `pnpm check`

– Vale

## Summary

- Preserves cross-tenant `searchTextCollections` hits by deduplicating on tenant and record ID.
- Restores the primary `tenantKey` with `finally` after tenant fan-out.
- Updates slot, manifest, host-flow, and assemble-normalization expectations.
- Adds coverage for duplicate record IDs across tenants.

## Review concerns

- The shared client still mutates `tenantKey` before an awaited RPC.
- Overlapping calls can interleave and use the wrong tenant.
- Restoring `tenantKey` in `finally` does not isolate tenant selection per request.
- Add a concurrency regression test.
- Run and report the required local repository checks.

## Complexity

- For `n` aggregated results, composite-key deduplication remains **O(n)** time and **O(n)** additional space.
- The `try/finally` change does not materially increase asymptotic complexity.
- Cyclomatic complexity increases because of the `try/finally` control-flow path and tenant/result guards.
- The concurrency issue is a correctness risk, not an asymptotic complexity regression.

### DIFF
--- a/src/libravdb-client.ts
+++ b/src/libravdb-client.ts
@@ -451,19 +451,23 @@ export class LibravDBClient {
     const savedKey = this.tenantKey;
     const allResults: SearchTextResponse["results"] = [];
     const seen = new Set<string>();
-    for (const t of this.readTenants) {
-      this.tenantKey = t;
-      try {
-        const resp = await this.client.searchTextCollections(req);
-        for (const r of resp.results ?? []) {
-          if (!seen.has(r.id)) {
-            seen.add(r.id);
-            allResults.push(r);
+    try {
+      for (const t of this.readTenants) {
+        this.tenantKey = t;
+        try {
+          const resp = await this.client.searchTextCollections(req);
+          for (const r of resp.results ?? []) {
+            const dedupeKey = `${t}\0${r.id}`;
+            if (!seen.has(dedupeKey)) {
+              seen.add(dedupeKey);
+              allResults.push(r);
+            }
           }
-        }
-      } catch { /* skip failed tenant reads */ }
+        } catch { /* skip failed tenant reads */ }
+      }
+    } finally {
+      this.tenantKey = savedKey;
     }
-    this.tenantKey = savedKey;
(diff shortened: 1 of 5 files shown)


compoodment:
## Vale Review — PR #364

**Quality:** Q4/5 — sharp
**Head:** 85cf576

**Findings:**
- src/libravdb-client.ts:456 — `searchTextCollections()` still swaps `this.tenantKey` on the shared client immediately before an awaited RPC, so two overlapping calls on the same `LibravDBClient` can interleave and send a request under the wrong tenant key. The `finally` restores the key after the loop, but it does not isolate tenant selection per request. major

**Proof gaps:** No local gate run; finding is from current head source and requires a concurrency regression test to pin down.

**Verdict:** request-changes — the dedupe fix is good, but tenant fan-out still routes through shared mutable state across await boundaries.

– Vale

PR #366, opened 2026-07-13
fix: merge dream and normal memory search

### LINKED ISSUE #347: Bug: Dream search replaces all normal memory search (dream-only branch skips user/global/session collections)
## Description

When `dreamQuery.active` is `true`, `searchMemories` in `src/memory-runtime.ts` (lines 119–125) uses a ternary that branches to **either** the dream collection **or** normal collections — never both:

```typescript
const result = dreamQuery.active && cfg.crossSessionRecall !== false && searchCorpus !== "sessions"
  ? await client.searchText({
      collection: resolveDreamCollection(userId),
      text: queryText,
      k,
    })
  : await searchResolvedCollections(client, cfg, userId, sessionId, queryText, k, searchCorpus, params.kind, params.signals);
```

When the dream path is taken, `searchResolvedCollections` is never called. This means user, global, and session collections return **zero results** for any query that matches a dream signal pattern.

Additionally, the `kind` and `signals` filter parameters are silently dropped on the dream path — `client.searchText` is called without them.

## Steps to Reproduce

1. Ingest memories into user/global/session collections
2. Send a query matching a dream signal pattern (e.g., "I had a dream about my childhood")
3. Observe that `searchMemories` returns only dream-collection results
4. If the dream collection is empty, the agent sees **zero** memory hits despite relevant non-dream memories existing

## Expected Behavior

Dream collection should be searched **in addition to** normal collections (merged/combined results), not **instead of** them.

## Impact

Any dream-recall query gets no results from normal memory. Combined with the false-positive over-filtering bug in `dream-routing.ts`, even idiomatic uses of "dream" (dream house, dream job) can suppress all memory search entirely.

## Fix Suggestion

Merge dream and normal search results rather than branching:

```typescript
const [normalResult, dreamResult] = await Promise.all([
  searchResolvedCollections(...),
  dreamQuery.active ? client.searchText({ collection: resolveDreamCollection(userId), ... }) : Promise.resolve(null),
]);
// merge results with dedup
```

compoodment:
## Summary

- Search the regular session/user/global collections for dream queries instead of replacing them with the dream collection.
- Search the dedicated dream collection alongside ordinary memory, then merge by score, deduplicate by record ID, and cap the combined result to the requested limit.
- Forward `kind` and `signals` filters to the dream search path.

## Root cause

Dream routing used an exclusive ternary: once a dream signal was active, it called only `searchText` for `dream:<user>` and skipped `searchResolvedCollections`. That suppressed all relevant ordinary memory and also dropped structured filters.

## Impact

Dream-related queries can now retrieve both promoted dream entries and related normal memories. An empty dream collection no longer makes an otherwise valid memory query return no hits.

## Verification

- `pnpm exec tsc --noEmit`
- `pnpm run clean:test && pnpm exec tsc -p tsconfig.tests.json && node --test .ts-build/test/unit/memory-runtime.test.js` — 18/18 passed
- `pnpm check` — reached the full unit suite; 218/220 passed, with two pre-existing failures in `slot-conflict.test.ts` caused by stale expected tool lists on `main`
- `git diff --check`

Fixes #347.

– Vale

### DIFF
--- a/src/memory-runtime.ts
+++ b/src/memory-runtime.ts
@@ -116,13 +116,35 @@ function createMemorySearchManager(
       const minScore = normalizeNumber(params.minScore);
       const client = await getClient();
 
-      const result = dreamQuery.active && cfg.crossSessionRecall !== false && searchCorpus !== "sessions"
-        ? await client.searchText({
-            collection: resolveDreamCollection(userId),
-            text: queryText,
+      const dreamCollection = dreamQuery.active && cfg.crossSessionRecall !== false && searchCorpus !== "sessions"
+        ? resolveDreamCollection(userId)
+        : undefined;
+      const normalSearch = searchResolvedCollections(
+        client,
+        cfg,
+        userId,
+        sessionId,
+        queryText,
+        k,
+        searchCorpus,
+        params.kind,
+        params.signals,
+      );
+      const result = dreamCollection
+        ? mergeSearchResults(
+            await Promise.all([
+              normalSearch,
+              client.searchText({
+                collection: dreamCollection,
+                text: queryText,
+                k,
+                kind: params.kind || undefined,
+                signals: params.signals && params.signals.length > 0 ? params.signals : undefined,
+              }),
+            ]),
             k,
-          })
-        : await searchResolvedCollections(client, cfg, userId, sessionId, queryText, k, searchCorpus, params.kind, params.signals);
+          )
+        : await normalSearch;
       const filteredResults =
         minScore === undefined
           ? result.results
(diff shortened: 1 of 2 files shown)


compoodment:
## Vale Review — PR #366

**Quality:** Q4/5 — sharp
**Head:** 2788016

**Findings:**
- src/memory-runtime.ts:255 — mergeSearchResults() deduplicates the dream and normal result streams by bare item.id, so a dream result and a normal/session/user result with the same record id collapse into one hit. This reintroduces the collection-collision class the repo already had to fix elsewhere, and it means a dream query can still hide relevant ordinary memory even though the PR is supposed to merge both paths. major

**Proof gaps:** No local gate run; needs a regression where dream:<user> and user:<user> or session:<id> return the same id with different text/collection and both survive the merge.

**Verdict:** request-changes — merge dedupe needs a collection-aware key, not bare record id.

– Vale
use @vale review

xDarkicex:
Superseded by merged PR #382, which combines this change with the complementary dream-search fix and includes full validation on OpenClaw 2026.8.2.

PR #367, opened 2026-07-13
fix: preserve dream signals in compound queries

### LINKED ISSUE #348: Bug: Dream false-positive filter kills entire dream signal on compound queries
## Description

In `src/dream-routing.ts` (lines 44–52), the false-positive check is all-or-nothing: if **any** false-positive pattern matches the query, the **entire** dream signal is suppressed — even when a genuine dream-recall pattern is also present:

```typescript
if (DREAM_FALSE_POSITIVE_PATTERNS.some((p) => p.test(queryText))) {
  return { active: false, matchedPatterns: [] };
}
```

## Steps to Reproduce

| Query | Expected | Actual |
|-------|----------|--------|
| "I had a dream about my dream house" | `active=true` (genuine dream recall) | `active=false` |
| "I had a dream about the American dream" | `active=true` (genuine dream recall) | `active=false` |
| "I had a dream last night about my dream job" | `active=true` (genuine dream recall) | `active=false` |

Each query contains a genuine dream-recall signal ("I had a dream") but also matches a false-positive pattern ("dream house", "American dream", "dream job"). The false-positive match kills the entire signal.

## Expected Behavior

False-positive patterns should only suppress the specific idiom match, not the entire dream-recall signal. A compound query with both genuine dream recall and an idiomatic use of "dream" should still activate dream search.

## Impact

Combined with issue #347 (dream search replaces normal search), these users get **zero memory results at all** — no dream search (killed by this bug) and no normal search (skipped by the other bug).

## Fix Suggestion

Instead of returning `{ active: false }` on any false-positive match, remove only the false-positive patterns from `matchedPatterns` and re-evaluate whether remaining genuine patterns still qualify. Or weight genuine vs. false-positive matches and only suppress if false-positives dominate.

compoodment:
## Summary

- Remove known idiomatic dream phrases from a query before evaluating genuine dream-recall patterns.
- Preserve dream routing when a real recollection phrase and an idiom appear in the same compound query.
- Keep false-positive-only queries such as “tell me about your dream job” on the ordinary memory path.

## Root cause

The detector rejected the entire query whenever any false-positive phrase appeared, even if another part of the same query independently matched explicit dream-recall intent.

## Impact

Queries such as “I had a dream about my dream house” now search dream memory, while idiomatic uses of “dream” alone remain filtered out.

## Verification

- `pnpm exec tsc --noEmit`
- `pnpm run clean:test && pnpm exec tsc -p tsconfig.tests.json && node --test .ts-build/test/unit/dream-routing.test.js` — 6/6 passed
- `git diff --check`

Fixes #348.

– Vale

## Summary

- Sanitizes idiomatic “dream” phrases before evaluating genuine dream-recall patterns.
- Preserves dream routing for compound queries while preventing idiom-only false positives.
- Adds coverage for mixed idiomatic/genuine queries and overlapping false-positive patterns.

## Verification

- TypeScript checks passed.
- Six dream-routing tests passed.
- `git diff --check` passed.

## Complexity

- **Big O:** No asymptotic regression; processing remains linear in query length for a fixed set of regex patterns.
- **Cyclomatic complexity:** No regression; per-rule false-positive branching was replaced with preprocessing and simplifies the detection flow.

### DIFF
--- a/src/dream-routing.ts
+++ b/src/dream-routing.ts
@@ -23,11 +23,11 @@ const DREAM_PATTERN_RULES: Array<{ label: string; patterns: RegExp[] }> = [
 
 /** Phrases that contain "dream" but are idiomatic (not memory-recall intent). */
 const DREAM_FALSE_POSITIVE_PATTERNS: RegExp[] = [
-  /\bpipe\s+dreams?\b/i,
-  /\bdream\s+team\b/i,
-  /\bamerican\s+dream\b/i,
-  /\bdream\s+(?:house|home|car|wedding|vacation|job|school)\b/i,
-  /\bin\s+(?:my|our|your)\s+dreams\b/i,
+  /\bpipe\s+dreams?\b/gi,
+  /\bdream\s+team\b/gi,
+  /\bamerican\s+dream\b/gi,
+  /\bdream\s+(?:house|home|car|wedding|vacation|job|school)\b/gi,
+  /\bin\s+(?:my|our|your)\s+dreams\b/gi,
 ];
 
 const DREAM_MATCHED_PATTERNS = ["dream"] as const;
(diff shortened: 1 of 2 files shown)


compoodment:
## Vale Review — PR #367

**Quality:** Q4/5 — sharp
**Head:** 45ac62b

**Findings:**
- src/dream-routing.ts:40 — None: idiomatic dream phrases are stripped before matching, so mixed real-recall queries still route while idiom-only queries do not. info

**Proof gaps:** No local gate run in this worker; PR reports TypeScript and focused dream-routing unit coverage.

**Verdict:** comment-only — self-authored PR, behavior is covered and no current-head blocker found.

– Vale
use @vale review

xDarkicex:
Superseded by merged PR #382, which combines this change with the complementary dream-search fix and includes full validation on OpenClaw 2026.8.2.

PR #368, opened 2026-07-13
fix: reset cooldown on circuit class change

### LINKED ISSUE #349: Bug: Circuit breaker cooldownUntil not reset on failure class change
## Description

In `src/context-engine.ts`, `trackBeforeTurnFailure` (around line 1711) resets `.class` and `.consecutive` when the failure class changes, but does **not** reset `.cooldownUntil`. This leaves a stale cooldown from the previous failure class in effect:

```typescript
function trackBeforeTurnFailure(sessionId: string, error: unknown): void {
  const cls = classifyError(error);
  let state = circuitBreakers.get(sessionId);
  if (!state) {
    state = { class: cls, consecutive: 0, lastFailure: 0, cooldownUntil: 0 };
  }
  if (state.class !== cls) {
    state.class = cls;
    state.consecutive = 0;        // reset
    // ⚠️ cooldownUntil is NOT reset — stale value persists
  }
  state.consecutive++;
  state.lastFailure = Date.now();
  const maxConsecutive = MAX_CONSECUTIVE_BEFORE_OPEN[state.class];
  if (state.consecutive >= maxConsecutive) {
    state.cooldownUntil = computeCooldown(state);  // only recomputed if threshold met
  }
  circuitBreakers.set(sessionId, state);
}
```

## Steps to Reproduce

1. Trigger 3 timeout errors → `class="timeout"`, `consecutive=3 >= 3`, `cooldownUntil = now + 15s` (circuit OPEN)
2. Trigger an unavailable error → `class` changes to `"unavailable"`, `consecutive` resets to 1
3. Since `1 < 2` (maxConsecutive for unavailable), `cooldownUntil` is **not recomputed**
4. `isBeforeTurnCircuitOpen` checks `Date.now() < cooldownUntil` → returns `true`
5. Circuit stays open based on the **timeout** class's 15s schedule, even though the failure mode has changed

## Expected Behavior

When the failure class changes, the circuit should either:
- Reset entirely (new class = fresh start, close the circuit), or
- Recompute `cooldownUntil` based on the new class's schedule

## Impact

A transient timeout followed by a different error type can leave the circuit breaker stuck open longer than intended, or with a backoff schedule from the wrong error class.

compoodment:
## Summary

- Clear `cooldownUntil` whenever a session’s BeforeTurn failure changes class.
- Add a regression that opens the timeout circuit, changes the failure to unavailable, and verifies the next call is not blocked by the stale timeout cooldown.

## Root cause

The failure-class transition reset `class` and `consecutive`, but retained the prior class’s `cooldownUntil`. Until the new class reached its own threshold, the old cooldown continued to make the circuit appear open.

## Impact

A new failure mode now starts with a clean circuit state instead of inheriting backoff timing from an unrelated earlier failure.

## Verification

- `pnpm exec tsc --noEmit`
- `pnpm run clean:test && pnpm exec tsc -p tsconfig.tests.json && node --test .ts-build/test/unit/context-engine.test.js` — 74/74 passed
- Regression confirmed to fail against the old implementation (`4 !== 5` BeforeTurn RPC calls)
- `git diff --check`

Fixes #349.

– Vale

### DIFF
--- a/src/context-engine.ts
+++ b/src/context-engine.ts
@@ -1809,6 +1809,7 @@ export function buildContextEngineFactory(
     if (state.class !== cls) {
       state.class = cls;
       state.consecutive = 0;
+      state.cooldownUntil = 0;
     }
     state.consecutive++;
     state.lastFailure = Date.now();
--- a/test/unit/context-engine.test.ts
+++ b/test/unit/context-engine.test.ts
@@ -577,6 +577,57 @@ test("context engine assemble strips OpenClaw untrusted metadata envelope from p
   assert.equal(call.params.prompt, "@User-1234 Reply with exactly PONG.");
 });
 
+test("context engine clears a stale circuit cooldown when the failure class changes", async () => {
+  class DeferredBeforeTurnClient extends FakeClient {
+    public pendingBeforeTurn: Array<{ reject(error: unknown): void }> = [];
+    public beforeTurnSucceeds = false;
+
+    beforeTurnKernel(params: Record<string, unknown>): Promise<{ predictions: [] }> {
+      this.calls.push({ method: "beforeTurnKernel", params });
+      if (this.beforeTurnSucceeds) return Promise.resolve({ predictions: [] });
+      return new Promise((_resolve, reject) => {
+        this.pendingBeforeTurn.push({ reject });
+      });
+    }
+  }
+
+  const client = new DeferredBeforeTurnClient();
+  const engine = buildContextEngineFactory(fakeRuntime(client), {
+    userId: "fixed-user",
+    beforeTurnTimeoutMs: 1000,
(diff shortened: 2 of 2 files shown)


compoodment:
## Vale Review — PR #368

**Quality:** Q4/5 — sharp
**Head:** ded601d

**Findings:**
- src/context-engine.ts:1808 — None: failure-class changes now clear the stale cooldown before counting the new class, so an unavailable failure no longer inherits a timeout cooldown. info

**Proof gaps:** No local gate run in this worker; PR reports TypeScript and a regression that fails against the old implementation.

**Verdict:** comment-only — self-authored PR, narrow fix, no current-head blocker found.

– Vale
use @vale review

PR #370, opened 2026-07-13
fix: sanitize multiline tool-call JSON

### LINKED ISSUE #353: Bug: TOOL_CALL_JSON_RE cannot match multiline JSON tool-call objects
## Description

In `src/context-engine.ts` (line 1151), `TOOL_CALL_JSON_RE` uses `[^\r\n]*` which cannot cross line boundaries. Multi-line JSON objects are never matched or sanitized:

```typescript
const TOOL_CALL_JSON_RE = /\{[^\r\n]*"name"\s*:\s*"([^"]+)"[^\r\n]*(?:"arguments"|"args"|"toolCallId"|"tool_call_id"|"type"\s*:\s*"toolCall")[^\r\n]*\}/g;
```

## Steps to Reproduce

Single-line: matched and sanitized ✓

Multiline: NOT matched, left in output ✗ — the regex cannot cross the newline boundaries inside the JSON object.

## Expected Behavior

Multi-line JSON tool-call objects should also be sanitized. Replace `[^\r\n]*` with `[\s\S]*?` (non-greedy any-character) — matching the pattern already used in `TOOL_CALL_BRACKET_RE` for its JSON alternation.

## Impact

Tool-call patterns in multi-line JSON survive sanitization and appear in compacted session context, potentially priming small models to repeat tool-call syntax.

compoodment:
## Summary

- Replace the single-line raw tool-call JSON regex with brace- and string-aware object scanning.
- Parse each complete candidate as JSON and remove it only when it contains a direct or nested tool-call record.
- Preserve adjacent ordinary JSON instead of allowing a multiline match to span unrelated objects.

## Root cause

`TOOL_CALL_JSON_RE` used `[^\r\n]*`, so it could never match a formatted JSON object across line breaks. Simply changing it to a cross-line wildcard would risk consuming from an earlier ordinary object through a later tool call, especially with nested arguments.

## Impact

Historical multiline tool-call JSON is now removed from recalled context before it can prime a model to repeat tool syntax, without deleting neighboring ordinary structured memory.

## Verification

- `pnpm exec tsc --noEmit`
- Full `.ts-build/test/unit/context-engine.test.js` — 74/74 passed
- Focused historical-tool, ordinary-JSON, and multiline-tool-JSON tests — 3/3 passed
- `git diff --check`

Fixes #353.

– Vale

### DIFF
--- a/src/context-engine.ts
+++ b/src/context-engine.ts
@@ -1240,8 +1240,105 @@ function escapeMemoryFactText(text: string): string {
 // Matches [tool:name] followed by optional whitespace and any trailing JSON object {...}, array [...], or string "..."
 const TOOL_CALL_BRACKET_RE = /\[tool:([^\]]+)\](?:\s*(?:\{[\s\S]*?\}|\[[\s\S]*?\]|".*?"))?/gi;
 
-// Matches raw JSON tool-call objects targeting a "name\" field
-const TOOL_CALL_JSON_RE = /\{[^\r\n]*"name"\s*:\s*"([^"]+)"[^\r\n]*(?:"arguments"|"args"|"toolCallId"|"tool_call_id"|"type"\s*:\s*"toolCall")[^\r\n]*\}/g;
+const TOOL_CALL_JSON_MARKER_KEYS = ["arguments", "args", "toolCallId", "tool_call_id"] as const;
+
+function containsJsonToolCall(value: unknown): boolean {
+  const pending: unknown[] = [value];
+
+  while (pending.length > 0) {
+    const candidate = pending.pop();
+    if (Array.isArray(candidate)) {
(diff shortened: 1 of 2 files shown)


compoodment:
## Vale Review — PR #370

**Quality:** Q4/5 — sharp
**Head:** a3aab9c

**Findings:**
- src/context-engine.ts:1273 — None: the new scanner is string/brace-aware, and src/context-engine.ts:1326-1340 removes only parsed objects that contain a tool-call record while preserving neighboring ordinary JSON. info

**Proof gaps:** No local gate run in this worker; PR reports TypeScript and the focused context-engine unit coverage.

**Verdict:** comment-only — self-authored PR, focused fix, no current-head blocker found.

– Vale
use @vale review

PR #373, opened 2026-08-10
fix: rename the plugin CLI root from 'memory' to 'libravdb' to stop colliding with core

### LINKED ISSUE #6: CLI namespace: plugin still registers root `memory`, colliding with core — 4 subcommands unreachable
The tool-name migration to the `libravdb_` prefix was correct, but the **CLI namespace was never migrated** — the plugin still registers the root command `memory`, which core now owns. On every gateway boot:

```
[plugins] cli command already registered: memory (memory-core)
  (plugin=libravdb-memory,
   source=.../node_modules/@xdarkicex/openclaw-memory-libravdb/dist/index.js)
```

This is logged at **error** level. The plugin itself still loads — only CLI registration is skipped (`registry-registrars-operations.ts` returns early from the registrar, it does not abort the plugin) — so it is easy to miss.

### Versions

| component | version |
|---|---|
| `@xdarkicex/openclaw-memory-libravdb` | 1.10.21 |
| OpenClaw | 2026.8.1 |
| `libravdbd` | 1.9.35 |
| platform | macOS 26.6.1, arm64 |

### What was already done right

`openclaw.plugin.json` declares the tool contracts as:

```
libravdb_memory_search   ← prefixed
libravdb_memory_get      ← prefixed
memory_describe, memory_expand, memory_grep   ← unprefixed
```

That split is exactly correct and I want to be clear it isn't what I'm reporting. Core defines `memory_search` and `memory_get`; it does **not** define `memory_describe`, `memory_expand` or `memory_grep`, so leaving those three unprefixed is right, not an oversight.

The gap is that the same migration was never applied to the CLI surface.

### What is still claimed

The plugin registers one root, `memory`, with seven subcommands (`dist/cli.js`, `registerMemoryCli` → `ensureCommand(program, "memory")`). Because the collision is at the root, the whole subtree is dropped:

| plugin subcommand | core equivalent | result |
|---|---|---|
| `index` | `memory index` — *Reindex memory files* | shadowed, different semantics |
| `search` | `memory search` — *Search memory files* | shadowed, different semantics |
| `status` | `memory status` — *Show memory search index status* | shadowed, different semantics |
| `dream-promote` | none | **unreachable** |
| `export` | none | **unreachable** |
| `flush` | none | **unreachable** |
| `journal` | none | **unreachable** |

Two distinct problems:

1. **Four commands have no invocation path at all.** `flush`, `journal`, `export` and `dream-promote` cannot be run. `dream-promote` matters in particular because `dreamPromotionEnabled` / `dreamPromotionDiaryPath` are live config options — the feature is configurable but its command is unreachable.

2. **Three commands silently resolve to a different subsystem.** `openclaw memory search` searches memory *files* via core; the plugin's would have searched the vector store. The user gets a plausible answer from the wrong implementation with no error.

### Suggested fix

Rename the CLI root to `libravdb`, matching the tool prefix — `openclaw libravdb search`, `openclaw libravdb flush`, and so on. That recovers the four dead commands and removes the three-way ambiguity in one move.

`activation.onCommands` in `openclaw.plugin.json` currently reads `["memory"]` and would need the same change, otherwise activation still keys off the core-owned root.

### Reproduction

```bash
openclaw plugins doctor 2>&1 | grep libravdb
# - libravdb-memory: cli command already registered: memory (memory-core)

openclaw memory --help    # shows core's subcommands; none of the plugin's
```

Marvinthebored:
Fixes xDarkicex/homebrew-openclaw-libravdb-memory#6

## Problem

OpenClaw core registers a `memory` CLI command. This plugin registers the same root, and core wins. Three consequences on a current host (measured on 2026.8.1):

1. **Four core subcommands are shadowed** — `openclaw memory index|promote|promote-explain|rem-backfill|rem-harness` were unreachable while this plugin was enabled.
2. **The plugin's own CLI is unreachable** — none of `status|index|search|flush|export|journal|dream-promote` can be invoked.
3. **Every gateway startup logs an error:**

```
[plugins] cli command already registered: memory (memory-core)
  (plugin=libravdb-memory, source=.../@xdarkicex/openclaw-memory-libravdb/dist/index.js)
```

## What this changes

Renames the CLI root to `libravdb` in **both** places that declare it — they have to agree, because the host uses the descriptor for collision detection and the builder for the actual command tree:

- `src/cli-descriptors.ts` — `MEMORY_CLI_DESCRIPTOR.name`
- `src/cli.ts` — `ensureCommand(program, …)`

and the three things that follow from it:

- `openclaw.plugin.json` — `activation.onCommands: ["memory"] → ["libravdb"]`, so lazy activation still triggers on the command that now exists
- `test/unit/cli.test.ts`, `test/integration/checklist-validation.test.ts` — the assertions that pin the root
- 33 references across `README.md`, `docs/*`, `POTENTIAL_BUGS.md` and `scripts/auto-install.sh`

**Deliberately unchanged**, because they are capability, slot and tool identifiers rather than the CLI namespace: `kind: ["memory", "context-engine"]`, `contracts.memory`, `plugins.slots.memory`, and the `memory_describe` / `memory_expand` / `libravdb_memory_*` tool names.

`libravdb` is a suggestion, not a demand — it matches `libravdbd`, the package name and the tool prefix. Rename it to whatever you prefer; the point is that the two declaration sites agree and neither is `memory`.

## ⚠️ This is a breaking CLI rename

Plugin commands move from `openclaw memory …` to `openclaw libravdb …`. Shell scripts, aliases, runbooks and monitoring checks need updating. A compatibility alias under `memory` is not viable, since that root is the collision being fixed.

Also worth coordinating: `scripts/auto-install.sh` now verifies with `openclaw libravdb status`. It installs the unversioned package, so if the script ships before the matching npm release it would install an older plugin that only registers `memory` and then fail verification. Pinning a minimum version there may be worth doing separately — I have not done it in this PR.

## Evidence

**Both declarations agree in the built bundle:**

```
$ tsc -p tsconfig.build.json && node scripts/bundle-entrypoint.mjs
$ grep -o 'MEMORY_CLI_DESCRIPTOR = {\s*name: "[a-z]*"' dist/index.js   → libravdb
$ grep -o 'ensureCommand(program, "[a-z]*")'            dist/index.js   → libravdb
$ grep -c 'ensureCommand(program, "memory")'            dist/index.js   → 0
```

**The startup collision is gone.** Installed the built bundle on a disposable OpenClaw 2026.8.1 profile with `plugins.slots.memory = libravdb-memory` and started a gateway:

```
$ grep -c "cli command already registered" gateway.log
0
LibraVDB registering mode=full
http server listening (14 plugins …)
```

Before this change the same startup logged the collision every time.

**Core's `memory` is fully unshadowed** — all core subcommands are reachable again with the plugin enabled:

```
$ openclaw memory --help
Usage: openclaw memory [options] [command]
Search, inspect, and reindex memory files
Commands:
  index             Reindex memory files
  promote           Rank short-term recalls and optionally append top entries to MEMORY.md
  promote-explain   Explain a specific promotion candidate and its score breakdown
  rem-backfill      Write grounded historical REM summaries into DREAMS.md for UI review
  rem-harness       Preview REM reflections, candidate truths, and deep …
```

**Tests:** `tsc --noEmit` clean; `test/unit/cli.test.ts` 15 pass / 0 fail; `checklist-validation` 0 fail.

⚠️ **Two `context-engine` unit tests fail — but they fail identically on unmodified `main`**, verified by stashing this change and re-running (89 pass / 2 fail both ways). They are unrelated to this PR.

## What I could not verify

I could **not** confirm `openclaw libravdb …` resolving end-to-end on my host. Reaching the plugin's CLI requires the `./dist/cli-metadata.js` entry, and with both entries listed my host's OpenClaw assigns the plugin an entry-scoped id (`libravdb-memory/index`), which breaks slot matching for unrelated reasons — so the plugin CLI was unreachable there before and after this change. What I have demonstrated is that the collision disappears, core is unshadowed, and both declaration sites now agree. If you can test the resolved command tree on a clean install, that is the gap worth closing.

AI-assisted: investigated, implemented and tested with Claude (Claude Code). Two adversarial review passes; the first found that `openclaw.plugin.json` and the tests were missed, which is why they are included.

## Summary

Renames the plugin CLI root from `memory` to `libravdb` to remove the collision with OpenClaw core.

- Updates CLI registration, activation, command building, tests, scripts, and documentation.
- Preserves capability identifiers, plugin slots, contracts, tool names, and subcommand behavior.
- Restores access to OpenClaw core `memory` commands.
- TypeScript checks and relevant tests pass.
- End-to-end `openclaw libravdb …` resolution remains unverified because of an unrelated entry-scoped plugin ID issue.

## Complexity

- Big O complexity: unchanged. The changes rename command strings and do not alter command-processing algorithms.
- Cyclomatic complexity: unchanged. No new branches or control-flow paths were added.

### DIFF
--- a/POTENTIAL_BUGS.md
+++ b/POTENTIAL_BUGS.md
@@ -114,7 +114,7 @@ last user message if it exists in the new slice.
 
 ## Diagnostic commands
 
-- `openclaw memory status` — current counts, profile, sidecar health
+- `openclaw libravdb status` — current counts, profile, sidecar health
 - `libravdbd status -v --json` — daemon health, cache stats, tenant registry
 - `libravdbd search -k 30 -tenant <key> "query"` — see what the DB has
 - `ls -lt ~/.openclaw/libravdb-manifests/*.manifest.json | head -5` — see
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ brew tap xDarkicex/homebrew-openclaw-libravdb-memory && brew install libravdbd &
 openclaw plugins install @xdarkicex/openclaw-memory-libravdb
 
 # 3. Verify
-openclaw memory status
+openclaw libravdb status
 ```
 
 > ⚡ **Done.** Your agent has persistent vector memory, identity tracking, and causal graph reasoning.
@@ -157,7 +157,7 @@ openclaw plugins list | grep libravdb
 Verify the service and plugin:
 
 ```bash
-openclaw memory status
+openclaw libravdb status
 ```
 
 Healthy output should show `Kernel=running`, stored memory counts, the active
@@ -402,13 +402,13 @@ Before exposing OpenClaw over remote channels, read [Security](./docs/security.m
 ## Operator Quick Refs
 
 ```bash
-openclaw memory status [--deep] [--json]
-openclaw memory index --force
-openclaw memory search "prior context"
-openclaw memory export --user-id <userId>
-openclaw memory flush --user-id <userId>
-openclaw memory journal --limit 50
-openclaw memory dream-promote --user-id <userId> --dream-file ~/DREAMS.md
+openclaw libravdb status [--deep] [--json]
+openclaw libravdb index --force
+openclaw libravdb search "prior context"
+openclaw libravdb export --user-id <userId>
+openclaw libravdb flush --user-id <userId>
+openclaw libravdb journal --limit 50
+openclaw libravdb dream-promote --user-id <userId> --dream-file ~/DREAMS.md
 ```
 
 ### memory kernel CLI (libravdbd v1.6.0+)
@@ -718,7 +718,7 @@ dream-related questions (e.g. "what did I dream about?").
 You can also promote manually without enabling the watcher:
 
 ```bash
-openclaw memory dream-promote --user-id <userId> --dream-file ~/DREAMS.md
+openclaw libravdb dream-promote --user-id <userId> --dream-file ~/DREAMS.md
 ```
 - **Embedding profiles** default to `nomic-embed-text-v1.5` with `bge-small-en-v1.5`
   fallback. See [Embedding profiles](./docs/embedding-profiles.md).
--- a/docs/features.md
+++ b/docs/features.md
@@ -105,7 +105,7 @@ Automatic diary watching:
 Manual run:
 
 ```bash
-openclaw memory dream-promote --user-id <userId> --dream-file ~/DREAMS.md
+openclaw libravdb dream-promote --user-id <userId> --dream-file ~/DREAMS.md
 ```
 
 Dream diary files must live under the operator's home directory or the configured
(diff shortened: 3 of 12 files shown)


compoodment (changes requested):
## Vale Review — PR #373

**Quality:** Q4/5 — sharp
**Head:** 59cda04

**Findings:**
- docs/uninstall.md:98 — The post-uninstall check now tells operators to run `openclaw libravdb status`, but `registerMemoryCli` returns before registering any plugin CLI when the memory slot is not selected (`src/cli.ts:104`), which is the expected state after removing the plugin. This check will fail as an unknown command instead of confirming uninstall state; use plugin-slot inspection plus a separate vector-service probe, or explicitly document unknown-command as expected. minor
- test/unit/cli.test.ts:104 — The CLI regression only asserts that `libravdb` exists; it never asserts that this plugin stopped registering a `memory` root. A future regression that registers both roots would pass the helper and reintroduce the collision. nit

**Proof gaps:** I did not run a host-level command-tree test for `openclaw libravdb` plus core `openclaw memory`; the PR body also calls that end-to-end resolution unverified.

**Verdict:** request-changes — The core rename is right, but the uninstall docs now prescribe a broken verification command.

– Vale
use @vale review

Marvinthebored:
Both findings addressed.

**@compoodment — the uninstall check.** You are right, and it was worse than it looked. Before the rename, `openclaw memory status` fell through to OpenClaw's *own* `memory` command and appeared to succeed, so it falsely confirmed uninstalls that had not happened. My blanket rename turned that silent false-pass into a loud unknown-command, which is more honest but still a bad instruction.

Rewritten to check the two things that actually indicate uninstall state — that `plugins.slots.memory` no longer names this plugin and it is gone from `openclaw plugins list`, and that the vector service endpoint refuses connections — with an explicit note that `openclaw libravdb` being unknown is the *expected* post-uninstall result, not a failure.

**@coderabbitai — the negative assertion.** Added: the CLI test now asserts no root named `memory` is registered, which is the actual point of the rename. The previous assertions only checked that `libravdb` exists, so both roots could have coexisted and the test would still have passed.

I have not added the host-level resolution test. On my host the plugin CLI is unreachable regardless of this change, because with both `openclaw.extensions` entries present the host assigns an entry-scoped plugin id (`libravdb-memory/index`) that breaks slot matching — an unrelated OpenClaw issue. That gap is stated in the PR description; if you can run it on a clean install it would close the last unverified piece.

Tests: cli 15 pass / 0 fail.

compoodment (approved):
## Vale Review — PR #373

**Quality:** Q4/5 — sharp
**Head:** 039d627

**Findings:**
- None

**Proof gaps:** I did not verify host-level `openclaw libravdb` command resolution; the PR body still calls that end-to-end path unverified because of an unrelated plugin-id issue.

**Verdict:** approve — Current head addresses the broken uninstall check and pins that the plugin no longer registers the core-owned `memory` root.

– Vale
use @vale review

PR #375, opened 2026-08-14
fix: stop afterTurn silently losing ingestion durability

Marvinthebored:
Depends on #374. This branch is stacked on it, so the diff shown here contains that PR's commit as well — **review only the second commit**, `fix: stop afterTurn silently losing ingestion durability`. Its tests need #374's state isolation to be reproducible. Rebasing once #374 lands will reduce this to a single commit.

## Problem

Two independent ways the queued ingestion task can fail to make a turn durable without saying so.

**1. A batch the daemon confirmed none of is treated as ingested.**

When the daemon returns a cursor whose `lastProcessedIndex` sits before the batch just sent, it has confirmed none of those messages. The reconciliation computes `ackCount = 0` and falls straight through its guard:

```ts
const ackCount = Math.max(0, confirmedIndex - startIndex + 1);
if (ackCount > 0) {
  // ...append to manifest
}
```

Nothing is appended, nothing is logged, and execution continues into the post-ingest steps. The manifest never advances, so the same messages are re-sent on every subsequent turn — indefinitely, with no diagnostic explaining why the session makes no progress.

**2. `bootstrap` orphans an ingestion that is still in flight.**

```ts
asyncIngestionQueues.delete(sessionId);
```

That entry is the per-session serialization point, and `bootstrap` can run while the previous turn is still ingesting. Dropping it lets the next `enqueueAsyncIngestion` start from a fresh `Promise.resolve()` and run alongside the task it just orphaned. Both then load the same manifest, compute the same overlap, and ingest the same messages.

The delete is not needed for cleanup either — entries already remove themselves in a `.finally()` once settled.

## What this PR does

1. Throws when `ackCount === 0`. The existing `catch` reports it exactly as it reports every other ingest failure, and it stops execution reaching the post-ingest best-effort work for a turn that was never stored.
2. Leaves the queue entry alone in `bootstrap`.

## Evidence

Each new test fails only against its own fix, so neither is passing for the wrong reason:

```
reverting only the ackCount fix    ->  only "afterTurn fails loudly when the daemon confirms none of the batch" fails
reverting only the bootstrap fix   ->  only "bootstrap does not orphan an ingestion that is still in flight" fails
both reverted                      ->  both fail
patched, full unit suite (23 files)->  0 failures
```

The unconfirmed-batch test also asserts an absence — `prewarmEmbeddingCache` must not run for a turn the daemon did not confirm:

```ts
assert.equal(
  client.calls.filter((c) => c.method === "searchTextCollections").length,
  0,
  "post-ingest work ran for a turn the daemon did not confirm",
);
```

The concurrency test is deterministic rather than timing-dependent: ordering is controlled by explicit unresolved promises, with `setImmediate` used only to let already-scheduled continuations run. It asserts maximum observed concurrency is 1; without the fix it is 2.

## Risk

The throw is the higher-risk half, so it was checked against every path where `ackCount` could legitimately be 0:

- **Seed ingest / first turn / empty manifest** — `startIndex` is explicitly `0`, so zero ACK requires the daemon to return `lastProcessedIndex < 0` alongside a non-empty tail hash, which does mean nothing was acknowledged.
- **Cursor-gap repair** — a missing or empty `manifestTailHash` takes the repair branch earlier and never reaches this code.
- **Legacy daemon** — no cursor in the response takes the optimistic-ACK branch; unreachable.
- **Heartbeat turns** — same protocol; no new messages is handled by preflight before ingestion.
- **Excluded sessions** — return before anything is queued.
- **Incremental ingest** — `startIndex` always derives from the manifest's last acknowledged index; `prePromptMessageCount` cannot move it once the manifest is non-empty.

## Notes

AI-assisted: written with Claude (Claude Code). Reviewed adversarially before filing.

### DIFF
--- a/src/context-engine.ts
+++ b/src/context-engine.ts
@@ -2738,7 +2738,13 @@ export function buildContextEngineFactory(
       predictiveContextCache.delete(sessionId);
       postToolRecallCache.delete(sessionId);
       turnCache.invalidateSession(sessionId);
-      asyncIngestionQueues.delete(sessionId);
+      // Deliberately NOT clearing asyncIngestionQueues here. The queue entry is
+      // the per-session serialization point, and bootstrap can run while an
+      // ingestion from the previous turn is still in flight. Dropping the entry
+      // lets the next enqueueAsyncIngestion start from a fresh Promise.resolve()
+      // and run concurrently with the task it just orphaned, so two tasks load
+      // the same manifest, compute the same overlap and both ingest. Entries
+      // remove themselves once settled, so there is nothing to clean up.
       const userId = resolveUserId({
         userIdOverride: args.userId,
         sessionKey: args.sessionKey,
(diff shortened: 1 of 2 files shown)


compoodment (approved):
## Vale Review — PR #375

**Quality:** Q4/5 — sharp
**Head:** 1dbef14

**Findings:**
- None

**Proof gaps:** I did not run the unit suite in this automation pass; review is based on the current diff, PR evidence, and the targeted regression coverage added in test/unit/context-engine.test.ts.

**Verdict:** approve — The unconfirmed-cursor path now fails visibly, and bootstrap no longer tears down the per-session serialization point while ingestion is in flight.

– Vale
use @vale review

Marvinthebored:
@coderabbitai review

This PR's automated review was rate-limited when it was filed and never ran. Requesting it now — the second commit is the one to review; the first belongs to #374, which this is stacked on.

PR #376, opened 2026-08-14
fix: stop an unwalkable ingestion root from blocking gateway readiness

Marvinthebored:
## Problem

A markdown ingestion root whose directory enumeration blocks takes down the whole gateway, not just ingestion.

`start()` awaits a scan of every configured root in turn, so a walk that never returns keeps the sidecar phase from completing. The host never reaches ready, and everything gated on readiness stays down with it — including the cron scheduler.

This is not hypothetical. On macOS, enumerating a root under `~/Library/Mobile Documents` (iCloud Drive) without file-provider consent for the calling binary blocks roughly 20 seconds per `open` in an uninterruptible syscall and then fails `EINTR`, which libuv retries indefinitely:

```
[markdown-ingest] watch unavailable for /Users/…/Obsidian_vault/Projects:
  EINTR: interrupted system call, watch '…'
```

...and then no `generic scan complete` line for that root, ever. Gateway readiness reports:

```
GET /ready → 503 {"ready":false,"failing":["startup-sidecars"],"uptimeMs":109231,
                  "eventLoop":{"degraded":false,"utilization":0.047}}
```

Note the event loop is healthy — the block sits on a libuv threadpool thread, so the process looks fine while never becoming ready. On the machine this was found, the scheduler was down for **48 hours** and restarting never helped, because the condition reproduces on every boot. Each blocked call also permanently occupies one of libuv's four default threadpool threads.

## What this PR does

**1. A stall watchdog, not a time limit.** The walk is bounded by `markdownIngestionWalkTimeoutMs` (default 60s), measured as *time since the last filesystem call returned*, tracked per root. A large or cold root that keeps completing work is never quarantined however long it takes overall; only a root where nothing at all completes for the interval is. On a stall the root is quarantined for the process lifetime — never scanned again, no new watchers — so the thread already lost is the last one. A restart retries it once.

**2. Nothing downstream of a partial walk runs.** `pruneDeletedFiles` retires every document whose file is missing from the walk's current-files set, so feeding it a partial set deletes documents for files that were merely never reached.

**3. Transient errors stop being treated as deletion evidence.** Three paths deleted documents from the daemon on any failure, and are now scoped to a confirmed `ENOENT` — checked via the error `code` through the cause chain, never by string-matching a message:

- `safeReadFileStreamed` returned `null` on *any* read failure, and the caller deletes on `null`. An `EPERM`/`EIO`/`EINTR` now keeps the document.
- A failed `stat` during the walk silently left the file out of current-files, so the prune pass deleted its document. A non-`ENOENT` stat failure now protects the file from pruning.
- A failed `readdir` was swallowed and the walk still counted as complete, so every file behind that directory was pruned. Such a walk is now marked partial: files found elsewhere still sync, pruning is skipped for that scan, and files that *were* seen are still tracked so a later complete scan can prune them properly.

Transient failures now also surface in `syncErrors` rather than disappearing into the skipped count.

## Evidence

Six new tests, each verified to fail against the specific defect it covers (removing just that fix, keeping the tests):

| test | fails without |
|---|---|
| hung root is quarantined instead of blocking `start()` | the watchdog — the suite times out outright |
| slow-but-progressing root is **not** quarantined | the stall detector (a total-elapsed cap quarantines it) |
| transient read failure keeps the document; ENOENT deletes it | ENOENT discrimination in the read path |
| transient stat failure protects the file from the prune pass | ENOENT discrimination in the walk |
| document behind an unreadable directory is not pruned | the partial-walk prune gate |
| obsidian note that stops qualifying is not deleted twice | retirement recording at the deletion site |

```
integration suite   28 passed, 0 failed
unit suite           0 failed
tsc --noEmit         clean
```

The quarantine test runs a healthy sibling root alongside the hung one and asserts the healthy root still ingests, that a document seeded under the hung root survives, and that the hung root is not re-enumerated afterwards.

## Known limitations

Stated plainly rather than papered over:

- **Startup delay is bounded, not eliminated.** `refresh()` still scans roots sequentially, so a stalled root delays readiness by up to one stall timeout, and N stalled roots by N times that. What changes is that the delay is now finite and self-healing instead of permanent: before this, one stalled root meant readiness never arrived at all, on every boot, forever. Operators with slow or remote roots can lower `markdownIngestionWalkTimeoutMs` to shrink the worst case. Making the initial scan fully concurrent would shrink it further, but `syncCandidates` shares backpressure state (`lastAcceptMore`, `lastRetryAfterMs`) across roots, so that is a separate change with its own review surface rather than something to bundle here.
- Quarantine lasts for the process lifetime; recovery needs a restart.
- The blocked filesystem call cannot be cancelled and may keep occupying a libuv worker until the process exits.
- A watcher registered before the stall stays open until `stop()`, though it can no longer schedule a scan.
- The watchdog covers directory walking, not later file reads, ingestion RPCs, or a synchronously blocking `watch()` call.
- Detection is poll-based, so quarantine can land somewhat later than the configured threshold depending on when the last completion was observed.

## Notes

AI-assisted: written with Claude (Claude Code). Reviewed adversarially before filing — eight passes, each of the first seven finding a real defect in the fix for the previous finding, including two cases where a regression test passed vacuously and had to be rebuilt around the shape that actually reaches the bug.

## Summary

- Add a per-root filesystem stall watchdog.
- Use `markdownIngestionWalkTimeoutMs`, with a default of 60 seconds.
- Quarantine stalled roots for the process lifetime.
- Skip scans and watchers for quarantined roots until restart.
- Treat walks with transient failures as partial.
- Skip downstream pruning after partial walks.
- Delete documents only when the error cause chain confirms `ENOENT`.
- Preserve documents after transient `read`, `stat`, or `readdir` failures.
- Preserve original errno details in transient-failure diagnostics.
- Track retired documents, including oversized and Obsidian-filtered files.
- Register and document the timeout in the plugin schema.
- Add integration coverage for stalled roots, quarantine behavior, transient failures, safe pruning, progress-aware timeouts, and document retirement.
- Reported validation passes 28 integration tests, with no unit-test failures and clean TypeScript compilation.

## Complexity review

- Normal directory traversal remains **O(F + D)** per root, where `F` is the number of files and `D` is the number of directories visited.
- Retirement and reconciliation tracking uses **O(F + D)** additional memory per scanned root.
- Timeout checks and error classification add **O(1)** work per filesystem operation.
- Quarantine prevents repeated scans of stalled roots during the process lifetime.
- No clear asymptotic time regression is identified from the provided change summary.
- Cyclomatic complexity increases in the walk, error classification, quarantine, partial-scan, and retirement paths.
- New branches handle timeouts, partial walks, `ENOENT`, transient errors, filtering, quarantine, and duplicate retirement.
- Regression tests cover the added branches and failure modes.

### DIFF
--- a/openclaw.plugin.json
+++ b/openclaw.plugin.json
@@ -444,6 +444,11 @@
         ],
         "default": "mtime"
       },
+      "markdownIngestionWalkTimeoutMs": {
+        "type": "number",
+        "default": 60000,
+        "description": "Per-root stall timeout for markdown ingestion, in milliseconds. Measured as time since the last filesystem call returned, NOT a budget for the whole walk: a large or slow root that keeps making progress is never interrupted. A root that goes this long without a single completed filesystem call is treated as blocked and quarantined for the lifetime of the process."
+      },
       "markdownIngestionMaxTokensPerFile": {
         "type": "number",
         "default": 128000
--- a/src/markdown-ingest.ts
+++ b/src/markdown-ingest.ts
@@ -454,8 +454,17 @@ class DirectoryMarkdownSourceAdapter implements MarkdownSourceAdapter {
             // still is: union what we did see into knownFiles, otherwise a
             // file first discovered during a partial walk is never tracked
             // and a later complete scan cannot prune it when it is deleted.
+            // Only files that still hold a document count -- syncing may have
+            // established a deletion (vanished between readdir and read, or
+            // grown oversized), and re-adding those would make the next
+            // complete scan delete an already-deleted document. Every deletion
+            // path clears fileStates, so it is the authority on what exists.
             for (const file of currentFiles) {
-              rootState.knownFiles.add(file);
+              if (this.fileStates.has(file)) {
+                rootState.knownFiles.add(file);
+              } else {
+                rootState.knownFiles.delete(file);
+              }
             }
           } else {
             await this.pruneDeletedFiles(rootState, currentFiles, stats);
--- a/src/markdown-ingest.ts
+++ b/src/markdown-ingest.ts
@@ -97,6 +97,13 @@ interface RootState {
    */
   walkCompletions: number;
   /**
+   * Paths whose document this scan retired. Recorded explicitly rather than
+   * inferred from fileStates, so the partial-walk reconciliation removes only
+   * what it actually deleted and never drops tracking for a file that merely
+   * has no local state yet.
+   */
+  retiredThisScan: Set<string>;
+  /**
    * Set when a directory in this walk could not be enumerated for a reason
    * other than ENOENT. The resulting file set is partial through no fault of
    * the filesystem's contents, so it must not drive the prune pass.
@@ -408,6 +415,7 @@ class DirectoryMarkdownSourceAdapter implements MarkdownSourceAdapter {
       knownFiles: this.snapshotFilesForRoot(resolved),
       walkCompletions: 0,
       walkIncomplete: false,
+      retiredThisScan: new Set<string>(),
       quarantined: false,
       directoryWatchers: new Map<string, FsWatcherLike>(),
     };
@@ -439,6 +447,7 @@ class DirectoryMarkdownSourceAdapter implements MarkdownSourceAdapter {
           const candidates: FileCandidate[] = [];
           rootState.walkCompletions = 0;
           rootState.walkIncomplete = false;
+          rootState.retiredThisScan.clear();
           const walked = await this.walkDirectoryWithTimeout(rootState, currentFiles, stats, candidates);
           if (!walked) {
             // The walk is still blocked in a syscall that may never return.
@@ -454,16 +463,17 @@ class DirectoryMarkdownSourceAdapter implements MarkdownSourceAdapter {
             // still is: union what we did see into knownFiles, otherwise a
             // file first discovered during a partial walk is never tracked
             // and a later complete scan cannot prune it when it is deleted.
-            // Only files that still hold a document count -- syncing may have
-            // established a deletion (vanished between readdir and read, or
-            // grown oversized), and re-adding those would make the next
-            // complete scan delete an already-deleted document. Every deletion
-            // path clears fileStates, so it is the authority on what exists.
+            // Files this scan retired (vanished between readdir and read, or
+            // grown past the per-file cap) must NOT be re-added, or the next
+            // complete scan would delete an already-deleted document.
+            // Everything else stays tracked, including candidates left unsynced
+            // by backpressure: those may still hold a document, and dropping
+            // them would orphan it.
             for (const file of currentFiles) {
-              if (this.fileStates.has(file)) {
-                rootState.knownFiles.add(file);
-              } else {
+              if (rootState.retiredThisScan.has(file)) {
                 rootState.knownFiles.delete(file);
+              } else {
+                rootState.knownFiles.add(file);
               }
             }
           } else {
@@ -697,6 +707,7 @@ class DirectoryMarkdownSourceAdapter implements MarkdownSourceAdapter {
         stats.filesDeferred++;
         if (await this.deleteCachedSourceDocument(candidate.path)) {
           stats.filesDeleted++;
+          rootState.retiredThisScan.add(candidate.path);
         }
         continue;
       }
@@ -706,6 +717,9 @@ class DirectoryMarkdownSourceAdapter implements MarkdownSourceAdapter {
           mtimeMs: candidate.mtimeMs,
           ctimeMs: candidate.ctimeMs,
         });
+        if (result === "deleted") {
+          rootState.retiredThisScan.add(candidate.path);
+        }
         recordSyncResult(stats, result);
       } catch (error) {
         stats.syncErrors++;
--- a/src/markdown-ingest.ts
+++ b/src/markdown-ingest.ts
@@ -11,6 +11,9 @@ import { IngestQueue } from "./ingest-queue.js";
 import type { ClientGetter } from "./plugin-runtime.js";
 
 const DEFAULT_DEBOUNCE_MS = 150;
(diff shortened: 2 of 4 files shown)


compoodment (commented):
## Vale Review — PR #376

**Quality:** Q3/5 — promising
**Head:** cac469f

**Findings:**
- openclaw.plugin.json:433 — The new `markdownIngestionWalkTimeoutMs` knob is only added to `PluginConfig`; the plugin schema still has `additionalProperties: false` and does not list the property near the other markdown ingestion settings, so normal plugin config cannot set the timeout this PR introduces. minor
- src/markdown-ingest.ts:375 — `refresh()` still awaits each root scan sequentially, and `scanRoot()` waits for `walkDirectoryWithTimeout` before returning. A stuck first scan is now bounded, not eliminated: default readiness can still be delayed by 60s per stalled root. info

**Proof gaps:** I did not run the integration suite or reproduce the filesystem stall on a live host; this is a diff/schema review against the current head.

**Verdict:** comment-only — The safety shape is much better, but the config/schema miss and bounded startup delay should be explicit before this is treated as a complete readiness fix.

– Vale
use @vale review

Marvinthebored:
Thanks both — the schema miss was a real defect, not a nit: with `additionalProperties: false`, setting `markdownIngestionWalkTimeoutMs` was a validation error, so the knob this PR introduces could only ever run at its built-in default. Fixed in 4bf8c40, registered next to its siblings with type `number` and default `60000`. Verified rather than assumed:

```
additionalProperties: False
unknown keys WITH fix   : []
unknown keys WITHOUT fix: ['markdownIngestionWalkTimeoutMs']
```

Both the schema description and the `PluginConfig` doc comment now lead with the thing most likely to be misread — it is a per-root **stall** timeout, not a budget for the whole walk.

@compoodment on the bounded startup delay: agreed, and I have made it explicit in the PR body rather than leaving it implied. `refresh()` does still scan roots sequentially, so a stalled root delays readiness by up to one timeout and N stalled roots by N times that. The claim I want to be precise about is narrower than "readiness is fixed": before this, one stalled root meant readiness never arrived at all, on every boot, permanently — the delay is now finite and self-healing, and operators with slow or remote roots can lower the timeout to shrink the worst case.

Making the initial scan concurrent would shrink it further, and I deliberately did not do it here: `syncCandidates` shares backpressure state (`lastAcceptMore`, `lastRetryAfterMs`) across roots, so parallel scans would interleave that state. That is a separate change with its own review surface, and bundling it would mean shipping a concurrency change under the cover of a safety fix.

CodeRabbit's three test-double nits were all correct and are in the same commit: `rmdir` was leaving descendant directories registered, the hang-prefix match would have caught a sibling named `hung-2`, and the slow-root test's 60ms-vs-100ms margin was tight enough to flake on a loaded runner — now 400ms, which still forces the walk to outlive a naive total-elapsed cap set at the same value, since that distinction is the whole point of the test.

Suite is 28 passing / 0 failing, and every test in the PR is verified by removing its specific fix and confirming the named assertion fires.

compoodment (commented):
## Vale Review — PR #376

**Quality:** Q3/5 — promising
**Head:** 4bf8c4

**Findings:**
- src/markdown-ingest.ts:375 — `refresh()` still awaits roots sequentially, so startup readiness is bounded rather than immediate: each stalled root can still cost one `markdownIngestionWalkTimeoutMs` window before quarantine. info
- src/markdown-ingest.ts:827 — transient stat/read errors are collapsed to generic thrown errors before the caller logs them, so the original errno detail is lost from scan diagnostics. minor

**Proof gaps:** I did not run the integration suite or reproduce the filesystem stall on a live host; this is a current-head diff review with PR metadata, comments, reviews, and check state inspected.

**Verdict:** comment-only — The schema miss from the prior Vale review is fixed, and the remaining issues are documented/bounded rather than merge blockers.

– Vale
use @vale review

Marvinthebored:
@compoodment the errno finding is fixed in fa3396d — and it was a better catch than "minor" suggests: errno text is literally how the incident behind this PR was diagnosed (the `EINTR` in a watch warning was the first thread pulled), so collapsing it out of the diagnostics was self-defeating.

The sentinels now carry `{ failed: formatError(error) }` captured where the error is still in hand; the thrown messages include it, and the walk's stat-failure branch — which previously logged nothing — now warns with the detail alongside the existing protect-from-prune behavior. No double logging: a walk-path stat failure never reaches `syncCandidates` (the file is excluded from candidates), and a sync-path failure has no walk warning.

Both tests now assert the errno survives to the log. On the previous head they fail with exactly this review's complaint:

```
expected a warning carrying the EPERM detail, got:
  ["[markdown-ingest] sync failed for …/note.md: read failed transiently for …/note.md; keeping existing document"]
expected a warning carrying the EINTR detail, got: []
```

Suite remains 28 passing / 0 failing. The sequential-`refresh()` info point stands as documented in the PR body — bounded, deliberate, and not bundled here for the backpressure-state reasons noted there.

compoodment (approved):
## Vale Review — PR #376

**Quality:** Q4/5 — sharp
**Head:** fa3396

**Findings:**
- src/markdown-ingest.ts:548 — The stall watchdog polls progress once per timeout window rather than rearming exactly when each filesystem call completes, so a syscall that hangs immediately after progress can be quarantined close to one extra window later than the schema wording implies; the PR body documents poll-based detection, and this is bounded rather than a correctness blocker. info
- test/integration/markdown-ingest.test.ts:1176 — Current regression coverage exercises hung-root quarantine, transient read/stat/readdir preservation, progress-aware slow walks, and Obsidian retirement; targeted TypeScript compilation plus those six markdown-ingestion tests pass on the PR head. info

**Proof gaps:** Full upstream gates were not run; I did not reproduce the live filesystem-provider stall outside the test double.

**Verdict:** approve — The prior schema and errno findings are fixed, deletion safety is covered by behavior tests, and the remaining startup delay is bounded and documented.

– Vale
use @vale review

Issue #378, opened 2026-08-26
bug: compacted projection is not recoverable on the next logical turn


Marvinthebored:
## Summary

With LibraVDB selected as the context engine and `ownsCompaction: true`, successful predictive
compaction reduces a long prompt from about 491k tokens to about 109k tokens during one logical turn.
The next logical turn can fail to recover the daemon-owned compacted projection and send the complete
raw transcript again. In the captured incident, the next provider request had a 511,940-token
prompt total: 483,332 uncached input tokens plus 28,608 cache-read tokens.

This is real provider-reported prompt usage, not a UI or accounting artifact.

## Environment

- OpenClaw: `2026.8.1` (`32337f39da43e941d564cef2b30c6c4cd1cd72d5`)
- Plugin: `@xdarkicex/openclaw-memory-libravdb` `1.10.21`
- Plugin upstream main checked: `c883cf755609023cfc5b15c3acb04f08ed4d7c2c`
- Daemon: `libravdbd 1.9.36`
- Node.js: `v26.5.0`
- Host: macOS Darwin 25.6.0 arm64
- Provider path: OpenRouter Chat Completions
- Context window: 1,048,576 tokens
- `compactSessionTokenBudget`: 32,000
- `compactionThresholdFraction`: 0.8
- `compactThreshold`: unset

No message text or identifying session values are included below. `<SESSION>` replaces one stable
session ID.

### First transition

Provider-reported prompt usage (`input + cacheRead`):

| Local time | Input | Cache read | Prompt total |
|---|---:|---:|---:|
| 11:19:29 | 4,146 | 486,976 | 491,122 |
| 11:19:54 | 44,177 | 65,088 | 109,265 |
| 11:30:28 | 121 | 102,208 | 102,329 |
| 11:32:09 | 483,332 | 28,608 | 511,940 |

The host and plugin logs around the reinflation show:

```text
11:31:16.916 LibraVDB bootstrap sessionId=<SESSION>
11:31:16.980 predictive compaction trigger currentTokenCount=31177 threshold=16000
11:31:17.400 daemon state lastCompactedTurn=0 tokenAccumulatorAfter=0 totalTurns=0 skippedNoNewTurns=false
11:31:17.405 activated daemon-owned compacted projection source=compact
11:31:17.407 predictive compaction completed
11:31:19.742 pre-prompt messages=105 historyTextChars=27752 systemPromptChars=162667

11:31:23.177 new Discord message received for the same session
[...]

PR #379, opened 2026-08-26
fix: preserve compacted context across logical turns

Marvinthebored:
Related: #378

AI-assisted: OpenAI Codex implemented and tested this change. Claude Code (personal Opus and Sonnet) performed independent adversarial reviews. Public incident evidence is in #378; local review logs contain machine paths and are not published.

## What Problem This Solves

The context engine is created again for each logical turn. After compaction, the previous engine held the compacted projection only in its own memory. A later turn could therefore lose that projection and replay the full session transcript. In the observed incident, provider input rose from about 102k tokens to 511,940 tokens in one turn.

## Why This Change Was Made

This change keeps the last valid compacted projection in registration-scoped, process-local state. It records only a complete daemon-generated `<compacted_session_context>` marker, its source boundary, and a bounded boundary signature. Fresh engines can restore that projection while per-session lifecycle tokens reject late work from reset, ended, or superseded runs.

Compacted authority now fails closed. An actual daemon compaction invalidates the prior snapshot and rotates a projection-generation token, fencing assemblies that still hold the older projection. A daemon no-op that confirms the session is already compacted retains the valid snapshot. Queued after-turn work captures its lifecycle token before entering the queue, so a reset cannot let stale work reactivate compacted state.

This is a plugin-side mitigation. It does not establish durable daemon recovery across process restarts, so the broader issue remains open.

## User Impact

Ordinary new turns and superseded heartbeat turns retain the compacted projection instead of unexpectedly resending the full session. Stale projections cannot be restored after a newer compaction or a reset. This avoids large input-token and cost spikes. No configuration change is required.

## Evidence

- The branch includes current upstream `main` at `c883cf7`.
- On the previous PR head, the cached-marker/markerless-assembly regression and the overlapping-assembly regression both failed (0/2): each incorrectly returned `assembled` using stale authority. The queued after-turn/reset regression also failed (0/1) by reactivating compacted projection state after reset.
- On `14f7ec1`, all three regressions pass (3/3). Actual compaction invalidates the cached snapshot, overlapping assemblies are projection-generation fenced, and queued pre-reset work cannot call compaction or reactivate state.
- Full context-engine unit suite: 99/99 passed.
- Integration suites for checklist validation, dream promotion, host flow, and Markdown ingestion: 49/49 passed.
- `tsc --noEmit`, test compilation, and the production build passed.
- Independent read-only Codex adversarial reviews found two additional races during development. Both were fixed, regression-tested, and re-reviewed; the final review returned `SAFE`.
- The documented `pnpm check` and `npm run test:integration` wrappers stopped before running tests because this machine's pnpm policy rejects ignored dependency build scripts (`@google/genai`, `esbuild`, `koffi`, `openclaw`, `protobufjs`, and `sharp`). The underlying TypeScript, unit, and integration commands above passed without changing that dependency policy.

## Summary

- Persist the latest valid daemon-generated compacted context in registration-scoped process-local state.
- Restore the compacted projection when a context engine is recreated.
- Validate compaction markers, source boundaries, boundary signatures, projection tokens, and lifecycle tokens.
- Clear shared state and local caches after invalid responses, boundary divergence, reset, session end, or superseded runs.
- Clear projection state from already-created engines.
- Preserve live-tool context and session suffixes during assembly.
- Add tests for persistence, races, divergence, incomplete markers, cleanup, and failed-closed behavior.
- TypeScript checks and focused and integration suites passed.
- Documented wrapper commands did not run tests because the machine’s pnpm policy rejects ignored dependency build scripts.
- The state is process-local and does not provide recovery across process restarts.

## Complexity review

- Projection validation, boundary-signature generation, restoration, and transcript assembly remain **O(n)** in the number of session messages or marker content.
- State lookup, lifecycle-token checks, projection-token checks, and cleanup remain **O(1)** per session, excluding transcript and cache processing.
- No clear Big O regression is evident.
- Cyclomatic complexity increases in `context-engine.ts`, especially in compaction validation, projection invalidation, restoration, fallback handling, and lifecycle fencing.
- The implementation adds branches for race handling, reset handling, boundary divergence, stale projections, errors, and cleanup.
- The change increases control-flow complexity and review effort without changing the expected asymptotic complexity.

### DIFF
--- a/src/context-engine.ts
+++ b/src/context-engine.ts
@@ -2887,7 +2887,7 @@ export function buildContextEngineFactory(
         };
       }
 
-      const assembleLifecycleToken = getProjectionLifecycleToken(sessionId);
+      let assembleLifecycleToken = getProjectionLifecycleToken(sessionId);
       let cachedCompactedProjection = compactedProjectionState.snapshots.get(sessionId);
       if (cachedCompactedProjection) {
         if (
@@ -2897,9 +2897,8 @@ export function buildContextEngineFactory(
             cachedCompactedProjection.sourceStartIndex,
           ) !== cachedCompactedProjection.boundarySignature
         ) {
-          compactedProjectionState.snapshots.delete(sessionId);
-          compactedProjectionSessions.delete(sessionId);
-          postToolRecallCache.delete(sessionId);
+          clearCompactedProjectionState(compactedProjectionState, sessionId);
+          assembleLifecycleToken = getProjectionLifecycleToken(sessionId);
           cachedCompactedProjection = undefined;
         }
       }
--- a/src/context-engine.ts
+++ b/src/context-engine.ts
@@ -1160,6 +1160,33 @@ function hasCompactedSessionContext(text: string): boolean {
   return HAS_COMPACTED_SESSION_CONTEXT_RE.test(text);
 }
 
+function extractCompactedSessionContext(text: string): string | undefined {
+  return text.match(COMPACTED_SESSION_CONTEXT_RE)?.join("\n") || undefined;
+}
+
+type CompactedProjectionSnapshot = {
+  context: string;
+  sourceStartIndex: number;
+  boundarySignature: string;
+};
+
+export type CompactedProjectionState = {
+  snapshots: Map<string, CompactedProjectionSnapshot>;
+  lifecycleTokens: Map<string, object>;
+};
+
+export function createCompactedProjectionState(): CompactedProjectionState {
+  return { snapshots: new Map(), lifecycleTokens: new Map() };
+}
+
+export function clearCompactedProjectionState(
+  state: CompactedProjectionState,
+  sessionId: string,
+): void {
+  state.snapshots.delete(sessionId);
+  state.lifecycleTokens.delete(sessionId);
+}
+
 function sanitizeDaemonSystemPromptAddition(text: string): string {
   return demoteDaemonAuthoredContextBlocks(
     sanitizeToolCallPatterns(canonicalizeCompactedSessionContextBlocks(text)),
@@ -1323,6 +1350,20 @@ export function normalizeKernelMessages(
       return normalized;
     })
     .filter((message) => message.role === "user" || message.content.trim().length > 0);
+}
+
+function buildProjectionBoundarySignature(
+  messages: OpenClawCompatibleMessage[],
+  sourceStartIndex: number,
+): string {
+  // ponytail: OpenClaw histories are append-only; use a durable transcript
+  // revision instead if the host ever permits arbitrary mid-prefix edits.
+  const boundary = messages.slice(Math.max(0, sourceStartIndex - 4), sourceStartIndex);
+  return createHash("sha256").update(JSON.stringify({ sourceStartIndex, boundary: boundary.map((message) => ({
+    role: message.role,
+    content: message.content,
+    id: typeof message.id === "string" ? message.id : undefined,
+  })) })).digest("hex");
 }
 
 function buildBeforeTurnSignature(messages: OpenClawCompatibleMessage[]): string | null {
@@ -1976,6 +2017,7 @@ export function buildContextEngineFactory(
   runtime: PluginRuntime,
   cfg: PluginConfig,
   logger: LoggerLike = console,
+  compactedProjectionState: CompactedProjectionState = createCompactedProjectionState(),
 ) {
   if (cfg?.optimizationMemoCacheSize !== undefined) {
     setOptimizationMemoCacheSize(cfg.optimizationMemoCacheSize);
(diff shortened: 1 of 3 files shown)


compoodment (changes requested):
## Vale Review — PR #379

**Quality:** Q4/5 — sharp
**Head:** 744088e

**Findings:**
- src/context-engine.ts:1182 — `clearCompactedProjectionState()` only deletes the shared snapshot/lifecycle maps, but the active engine still owns `compactedProjectionSessions` and `postToolRecallCache`. A reset/session-end racing with an existing engine can therefore leave the old compacted-session activation and cached system prompt reachable through the post-tool shortcut at `src/context-engine.ts:3070`, so the lifecycle cleanup is not actually complete. major

**Proof gaps:** I did not rerun the local suites; GitHub only reports the CodeRabbit status on this head.

**Verdict:** request-changes — reset/session-end cleanup still leaves engine-local compacted projection state alive in the raced engine.

– Vale
use `@vale review`

Marvinthebored:
Addressed the requested lifecycle cleanup in b7eba7c. Compacted activation now lives in registration-scoped state, and clearCompactedProjectionState removes the snapshot, lifecycle token, activation, and post-tool cache entry, including for an already-created engine. The new same-engine reset regression proves a fresh daemon assembly and stale-marker absence; focused suites pass 106/106 and integration passes 49/49. A follow-up personal Claude Opus adversarial review returned SAFE after independently testing both halves of the fix. @vale review

Marvinthebored:
Addressed the overlapping-boundary race in 32cc60d. Boundary divergence now invalidates the old lifecycle generation and immediately captures a new token for the current call, so a delayed old RPC cannot overwrite the fresh projection and the current call is not self-fenced. The regression fails on b7eba7c (old call returns assembled), passes on 32cc60d, and asserts fresh shared state plus stale authority, prompt, and message absence. Focused suites pass 107/107; integration passes 49/49; final Claude Sonnet and read-only Codex reviews are SAFE. @vale review

compoodment (changes requested):
## Vale Review — PR #379

**Quality:** Q4/5 — sharp
**Head:** 32cc60d

**Findings:**
- src/context-engine.ts:2732 — `runCompaction()` still treats `result.compacted=true` as activation-only and leaves any existing shared snapshot untouched. If a session already has a cached marker, a later predictive compaction can succeed without returning a complete marker; the subsequent assembly path then restores the older marker at `src/context-engine.ts:3213` and reports `assembled` instead of failing closed. This can serve stale compacted context and stale source boundaries after a newer daemon compaction. major

**Proof gaps:** I did not rerun the local suites; current tests cover no-marker compaction only without an existing cached marker, so the cached-marker replacement case is still unproven.

**Verdict:** request-changes — successful compaction must clear or replace the prior snapshot before markerless assembly can restore obsolete state.

– Vale
use `@vale review`

Marvinthebored:
Addressed the stale-snapshot finding in 14f7ec1. Actual daemon compactions now invalidate the prior snapshot and rotate the projection generation; already-compacted no-op handoffs retain a valid snapshot. I also fenced overlapping assemblies and queued pre-reset after-turn compaction. Stock regressions failed 0/3; patched regressions pass 3/3. Full context-engine unit suite passes 99/99, integration passes 49/49, and typecheck/test compilation/build are clean. Final read-only Codex re-review returned SAFE. @vale review

compoodment (changes requested):
## Vale Review — PR #379

**Quality:** Q4/5 — sharp
**Head:** 14f7ec1

**Findings:**
- src/context-engine.ts:3616 — The queued `afterTurn` lifecycle token only gates `performAfterTurnPredictiveCompaction()` at `src/context-engine.ts:3696`; stale pre-reset work still calls `afterTurnKernel()`, updates the continuity cache, and can save the manifest at `src/context-engine.ts:3661`, `src/context-engine.ts:3683`, or `src/context-engine.ts:3693` after `clearCompactedProjectionState()` has invalidated the session. The new regression at `test/unit/context-engine.test.ts:1787` only asserts no `compactSession` call and never asserts that the stale queued ingest/manifest side effects are fenced, so reset/session-end can still be repopulated by old work before the new token check runs. major

**Proof gaps:** I did not rerun the local suites; this is a current-head source review against `14f7ec1` and the existing regression shape.

**Verdict:** request-changes — lifecycle fencing needs to happen before queued after-turn daemon/manifest side effects, not only before predictive compaction.

– Vale
use `@vale review`

Marvinthebored:
Addressed the queued after-turn lifecycle race in `16b2168`.

Queued ingestion now verifies the captured lifecycle token before it starts, after client acquisition, after each daemon response, and after predictive compaction. A reset therefore prevents waiting work from calling the daemon, and an already in-flight result cannot update the continuity cache, manifest, prediction cache, or projection state afterward.

The regression now covers both halves: one RPC already in flight and a second task waiting behind it. On `14f7ec1`, it fails because both stale tasks call `afterTurnKernel` (`2 !== 1`). On `16b2168`, only the already-started call is observed and the manifest remains empty.

Verification:

- focused stock regression: fails as expected
- focused patched regression: 1/1 pass
- context-engine unit suite: 99/99 pass
- integration suite: 49/49 pass
- TypeScript no-emit check: pass
- plugin inspector: pass, 0 breakages
- production build: pass

@vale review

compoodment (changes requested):
## Vale Review — PR #379

**Quality:** Q4/5 — sharp
**Head:** 16b2168

**Findings:**
- src/context-engine.ts:3827 — `dispose()` clears the engine-local caches and `asyncIngestionQueues`, but it never invalidates `compactedProjectionState.lifecycleTokens`. If disposal times out while an `afterTurn` task is already inside `afterTurnKernel()`, that stale task can resume with the same shared lifecycle token from `src/context-engine.ts:3575` and pass every `lifecycleIsCurrent()` check at `src/context-engine.ts:3578`, then save manifest state at `src/context-engine.ts:3668`, `src/context-engine.ts:3690`, or `src/context-engine.ts:3700` and run predictive compaction after the factory has supposedly been disposed. major

**Proof gaps:** I did not rerun local suites; this is a current-head source review. GitHub only reports CodeRabbit passing, and the new reset regression covers `clearCompactedProjectionState()` but not dispose timeout/factory teardown.

**Verdict:** request-changes — queued work that survives `dispose()` needs a factory-generation or shared lifecycle invalidation fence before it can write manifests or reactivate projection state.

– Vale
use `@vale review`

